PR #4, opened 2020-12-10
8258060: Update @jls tags for renamed/renumbered sections

jddarcy:
Given upcoming changes in the JLS terminology around the term "type", various sections were renamed:

    https://download.java.net/java/early_access/jdk16/docs/specs/class-terminology-jls.html

The @jls tags in the java.compiler module which refer to the renamed sections should be updated.

---------
### Progress

### Issue
 * [JDK-8258060](https://bugs.openjdk.java.net/browse/JDK-8258060): Update @jls tags for renamed/renumbered sections

### Reviewers
 * [Jonathan Gibbons](https://openjdk.java.net/census#jjg) (@jonathan-gibbons - **Reviewer**) ⚠️ Review applies to 9985bdbde9be32739cd774dec75d793aad8832ed
 * abuckley - Committer ⚠️ Added manually

### Download
`$ git fetch https://git.openjdk.java.net/jdk16 pull/4/head:pull/4`
`$ git checkout pull/4`

### DIFF
--- a/src/java.compiler/share/classes/javax/lang/model/AnnotatedConstruct.java
+++ b/src/java.compiler/share/classes/javax/lang/model/AnnotatedConstruct.java
@@ -129,10 +129,10 @@ import javax.lang.model.type.*;
  * </ul>
  *
  * @since 1.8
- * @jls 9.6 Annotation Types
+ * @jls 9.6 Annotation Interfaces
  * @jls 9.6.4.3 {@code @Inherited}
  * @jls 9.7.4 Where Annotations May Appear
- * @jls 9.7.5 Multiple Annotations of the Same Type
+ * @jls 9.7.5 Multiple Annotations of the Same Interface
  */
 public interface AnnotatedConstruct {
     /**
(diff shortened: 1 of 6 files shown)


src/java.compiler/share/classes/javax/lang/model/AnnotatedConstruct.java
```diff
@@ -184,7 +184,7 @@
      * @see IncompleteAnnotationException
      * @see MirroredTypeException
      * @see MirroredTypesException
-     * @jls 9.6.1 Annotation Type Elements
+     * @jls 9.6.1 Annotation Elements
```
jonathan-gibbons: Slightly surprising... but I agree that it matches JLS

jonathan-gibbons (approved):

src/java.compiler/share/classes/javax/lang/model/AnnotatedConstruct.java
jddarcy: Alex confirmed off-list that the intended title for that section is "Annotation Interface Elements"; I'll make the adjustment and get the change integrated.
Thanks.

PR #6, opened 2020-12-11
8257919: [JVMCI] profiling info didn't change after reprofile

iwanowww:
JDK-8252049 moved compiler counter initialization into MethodData constructor, but JVMCI and WhiteBox API rely on `MethodData::initialize()`/`MethodData::init()` to reset MDO state. 

Proposed fix restores the logic which resets the counters in `init()` (which is called from `initialize()`).

Testing:
- [x] hs-precheckin-comp, hs-tier1, hs-tier2

---------
### Progress

### Issue
 * [JDK-8257919](https://bugs.openjdk.java.net/browse/JDK-8257919): [JVMCI] profiling info didn't change after reprofile

### Reviewers

### Download
`$ git fetch https://git.openjdk.java.net/jdk16 pull/6/head:pull/6`
`$ git checkout pull/6`

### DIFF
--- a/src/hotspot/share/ci/ciMethodData.cpp
+++ b/src/hotspot/share/ci/ciMethodData.cpp
@@ -47,6 +47,7 @@ ciMethodData::ciMethodData(MethodData* md)
   _saw_free_extra_data(false),
   // Initialize the escape information (to "don't know.");
   _eflags(0), _arg_local(0), _arg_stack(0), _arg_returned(0),
+  _creation_mileage(0),
   _current_mileage(0),
   _invocation_counter(0),
   _backedge_counter(0),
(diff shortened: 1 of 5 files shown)


cl4es:
I guess the compiler will outsmart the back-to-back initialization but for clarity maybe the `MethodData` constructor should now initialize `_compiler_counters` using the empty constructor?

iwanowww:
After thinking more about it, I decided to revert `creation_mileage` change and move it back to MethodData: it turns out WhiteBox API deliberately keeps it intact while resetting other compiler-related counters (by calling `MethodData::init()` which was deliberately separated from `MethodData::initialize()` in JDK-8007288).  Now it should be back to pre-8252049 state and keeps the fix focused on native leak. The `MDO`/`ciMD` implementations can be refactored later.

iwanowww:
> I guess the compiler will outsmart the back-to-back initialization but for clarity maybe the MethodData constructor should now initialize _compiler_counters using the empty constructor?

Good point. Latest commit addresses that.

vnkozlov (approved):
Good. I verified locally that failed test passed.

cl4es (approved):

iwanowww:
Thanks for the reviews, Vladimir and Claus.

PR #7, opened 2020-12-11
8255381: com/sun/jdi/EATests.java should not suspend graal threads

reinrich:
This is a clone of https://github.com/openjdk/jdk/pull/1625 which was reviewed but not integrated before RDP1

The change is a test bug fix which can be integrated during RDP1 according to https://openjdk.java.net/jeps/3

--- Original Synopsis

This fixes a bug in the test test/jdk/com/sun/jdi/EATests.java that caused
timeout failures when graal is enabled.

The fix is to avoid suspending all threads when a breakpoint is reached and then resume
just the main thread again. This pattern was used in the test case
EAMaterializeLocalAtObjectPollReturnReturn. It caused timeouts because graal
threads remained suspended and, running with -Xbatch, the main thread waited
(with timeout) for completion of compile tasks.
The fix was applied to all breakpoints in the test. All explicit suspend calls now apply only
to the main test thread and all explicit resume calls apply to all java threads.

Testing: duration of the test case EAMaterializeLocalAtObjectPollReturnReturn is
reduced from 30s to 10s.

---------
### Progress

### Issue
 * [JDK-8255381](https://bugs.openjdk.java.net/browse/JDK-8255381): com/sun/jdi/EATests.java should not suspend graal threads

### Reviewers
 * [Chris Plummer](https://openjdk.java.net/census#cjplummer) (@plummercj - **Reviewer**)
 * [Martin Doerr](https://openjdk.java.net/census#mdoerr) (@TheRealMDoerr - **Reviewer**)
 * [Serguei Spitsyn](https://openjdk.java.net/census#sspitsyn) (@sspitsyn - **Reviewer**)

### Download
`$ git fetch https://git.openjdk.java.net/jdk16 pull/7/head:pull/7`
`$ git checkout pull/7`

### DIFF
--- a/test/jdk/com/sun/jdi/EATests.java
+++ b/test/jdk/com/sun/jdi/EATests.java
@@ -2109,7 +2109,7 @@ class EARelockingObjectCurrentlyWaitingOn extends EATestCaseBaseDebugger {
             inWait = env.targetMainThread.frame(0).location().method().name().equals("wait");
             if (!inWait) {
                 msg("Target not yet in java.lang.Object.wait(long).");
-                env.vm().resume();
+                env.targetMainThread.resume();
             }
         } while(!inWait);
         StackFrame testMethodFrame = env.targetMainThread.frame(4);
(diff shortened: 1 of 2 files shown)


reinrich:
@plummercj, @sspitsyn, I would like to bring this test bug fix to jdk16. It is identical to openjdk/jdk#1625.

Thanks, Richard.

plummercj (approved):

TheRealMDoerr (approved):

sspitsyn (approved):

reinrich:
Thanks for the reviews @plummercj, @sspitsyn, @TheRealMDoerr!

/integrate

PR #16, opened 2020-12-13
JDK-8247994: Localize javadoc search

jonathan-gibbons:
This is for JDK16, as a precursor to fixing JDK-8258002.

While it is good to be using localized strings in the generated output, the significance for JDK-8258002 is that the strings are now obtained from a resource file, and not hardcoded in JavaScript file itself.

The source file `search.js` is renamed to `search.js.template`, and (unlike other template files) is copied as-is into the generated image. The values in the template are resolved when javadoc is executed, depending on the locale in use at the time. Because of the change in the file's extension, two makefiles are updated to accommodate the new extension: one is for the "interim" javadoc used to generate the API docs; the other is for the primary javadoc in the main JDK image.

---------
### Progress

### Issue
 * [JDK-8247994](https://bugs.openjdk.java.net/browse/JDK-8247994): Localize javadoc search

### Reviewers
 * [Hannes Wallnöfer](https://openjdk.java.net/census#hannesw) (@hns - **Reviewer**)
 * [Magnus Ihse Bursie](https://openjdk.java.net/census#ihse) (@magicus - **Reviewer**) ⚠️ Review applies to 8eb7f27b703c1d5d713d5205145d3e9c856056ce

### Download
`$ git fetch https://git.openjdk.java.net/jdk16 pull/16/head:pull/16`
`$ git checkout pull/16`

### DIFF
--- a/make/CompileInterimLangtools.gmk
+++ b/make/CompileInterimLangtools.gmk
@@ -74,7 +74,7 @@ define SetupInterimModule
       EXCLUDE_FILES := $(TOPDIR)/src/$1/share/classes/module-info.java \
           Standard.java, \
       EXTRA_FILES := $(BUILDTOOLS_OUTPUTDIR)/gensrc/$1.interim/module-info.java, \
-      COPY := .gif .png .xml .css .js .txt javax.tools.JavaCompilerTool, \
+      COPY := .gif .png .xml .css .js .js.template .txt javax.tools.JavaCompilerTool, \
       BIN := $(BUILDTOOLS_OUTPUTDIR)/interim_langtools_modules/$1.interim, \
       DISABLED_WARNINGS := module options, \
       JAVAC_FLAGS := \
--- a/src/jdk.javadoc/share/classes/jdk/javadoc/internal/doclets/formats/html/resources/search.js.template
+++ b/src/jdk.javadoc/share/classes/jdk/javadoc/internal/doclets/formats/html/resources/search.js.template
@@ -37,7 +37,7 @@ var RANKING_THRESHOLD = 2;
 var NO_MATCH = 0xffff;
 var MIN_RESULTS = 3;
 var MAX_RESULTS = 500;
-var UNNAMED = "##REPLACE:doclet.search.unnamed##";
+var UNNAMED = "<Unnamed>>";
 function escapeHtml(str) {
     return str.replace(/</g, "&lt;").replace(/>/g, "&gt;");
 }
--- a/src/jdk.javadoc/share/classes/jdk/javadoc/internal/doclets/formats/html/resources/search.js.template
+++ b/src/jdk.javadoc/share/classes/jdk/javadoc/internal/doclets/formats/html/resources/search.js.template
@@ -23,13 +23,13 @@
  * questions.
  */
 
-var noResult = {l: "No results found"};
-var loading = {l: "Loading search index..."};
-var catModules = "Modules";
-var catPackages = "Packages";
-var catTypes = "Types";
-var catMembers = "Members";
-var catSearchTags = "SearchTags";
+var noResult = {l: "##REPLACE:doclet.search.no_results##"};
+var loading = {l: "##REPLACE:doclet.search.loading##"};
+var catModules = "##REPLACE:doclet.search.modules##";
+var catPackages = "##REPLACE:doclet.search.packages##";
+var catTypes = "##REPLACE:doclet.search.types##";
+var catMembers = "##REPLACE:doclet.search.members##";
+var catSearchTags = "##REPLACE:doclet.search.search_tags##";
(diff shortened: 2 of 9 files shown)


make/CompileInterimLangtools.gmk
```diff
@@ -74,7 +74,7 @@ define SetupInterimModule
       EXCLUDE_FILES := $(TOPDIR)/src/$1/share/classes/module-info.java \
           Standard.java, \
       EXTRA_FILES := $(BUILDTOOLS_OUTPUTDIR)/gensrc/$1.interim/module-info.java, \
-      COPY := .gif .png .xml .css .js .txt javax.tools.JavaCompilerTool, \
+      COPY := .gif .png .xml .css .js .js.template .txt javax.tools.JavaCompilerTool, \
```
jonathan-gibbons, line 77: Build-folk: it would be nice if this macro could use `$(jdk.javadoc_COPY)` instead of having to duplicate entries.
(Future RFE?)

src/jdk.javadoc/share/classes/jdk/javadoc/internal/doclets/formats/html/resources/search.js.template
```diff
@@ -23,21 +23,21 @@
...
+var catSearchTags = "##REPLACE:doclet.search.search_tags##";
 var highlight = "<span class=\"result-highlight\">$&</span>";
 var searchPattern = "";
 var fallbackPattern = "";
 var RANKING_THRESHOLD = 2;
 var NO_MATCH = 0xffff;
 var MIN_RESULTS = 3;
 var MAX_RESULTS = 500;
-var UNNAMED = "<Unnamed>";
+var UNNAMED = "##REPLACE:doclet.search.unnamed##";
```
hns: `<Unnamed>` is not a string that is ever shown to the user, it is what is used by javadoc to denote the default package (see `toolkit.util.DocletConstants.DEFAULT_PACKAGE_NAME`). This variable shouldn't be localized as that would break behaviour for code in the default package (and show the localized string as package name, instead of no package name).

src/jdk.javadoc/share/classes/jdk/javadoc/internal/doclets/toolkit/util/DocFile.java
```diff
@@ -216,6 +239,32 @@ public void copyResource(DocPath resource, boolean overwrite, boolean replaceNew
...
+        int start = 0;
+        while (m.find()) {
+            if (sb == null) {
+                sb = new StringBuilder();
+            }
+            sb.append(line, start, m.start());
+            try {
+                sb.append(resources.getText(m.group("key")));
+            } catch (MissingResourceException e) {
+                sb.append(m.group());
```
hns, line 256: Shouldn't we propagate an error here, or issue a warning? If a key is missing localized properties the value from default properties should be used, so this should never happen, right?

hns (changes requested):
Looks good in general. `<Unnamed>` shouldn't be localized as it is an internal tag (see inline comment).

src/jdk.javadoc/share/classes/jdk/javadoc/internal/doclets/toolkit/util/DocFile.java
hns: I see the added check for "##REPLACE:" in TestSearch.java will catch that case, so I guess it's ok.

make/CompileInterimLangtools.gmk
magicus: I agree. The entire design of CompileJavaModules.gmk needs to be updated. I've been procrastinating on cleaning this up, maybe it's time to get going on it...

magicus (approved):
Build changes look good.

src/jdk.javadoc/share/classes/jdk/javadoc/internal/doclets/formats/html/resources/search.js.template
jonathan-gibbons: Noted, thanks

src/jdk.javadoc/share/classes/jdk/javadoc/internal/doclets/formats/html/resources/search.js.template
```diff
@@ -37,7 +37,7 @@ var RANKING_THRESHOLD = 2;
 var NO_MATCH = 0xffff;
 var MIN_RESULTS = 3;
 var MAX_RESULTS = 500;
-var UNNAMED = "##REPLACE:doclet.search.unnamed##";
+var UNNAMED = "<Unnamed>>";
```
hns: Oops, an additional closing angle bracket slipped there.

hns (approved):

PR #20, opened 2020-12-15
JDK-8258002: Update "type" terminology in generated docs

jonathan-gibbons:
This is the work to update javadoc to use certain terminology dependent on the source version in use.

The groups of terms being changed are:

* `Types` to `Classes and Interfaces`
* `Annotation Type` to `Annotation Interface`
* `Enum` to `Enum Class`

The new terminology will be used when the source version of the elements being documented is 16 or higher. This is the default behavior. The corollary is that the old terminology will be used when the `--source` or `--release` option is used to specify a version less than `16`.

In addition, the terminology for `Record` is being unconditionally changed to use `Record Class`. Since the feature was in preview before this release, there is no need to be able to generate the previous form.

Also, in the course of this work, it was noted that on the "Use" page for an annotation type/interface, it was incorrectly classified as a *class*, instead of *annotation interface*: that has been fixed.

Fundamentally, the change is managed at a very low level, by passing a _mapping function_ into `Resource.getText`. This allows high level code, `HtmlDoclet` in this case, to provide version-specific overrides of the resource key to be used. The avoids having to have version-specific checks at the point of use for each affected resource.

---------
### Progress

### Issue
 * [JDK-8258002](https://bugs.openjdk.java.net/browse/JDK-8258002): Update "type" terminology in generated docs

### Reviewers
 * [Hannes Wallnöfer](https://openjdk.java.net/census#hannesw) (@hns - **Reviewer**) ⚠️ Review applies to 862ba59b248e9420e311832e308c5bc1352800bc

### Download
`$ git fetch https://git.openjdk.java.net/jdk16 pull/20/head:pull/20`
`$ git checkout pull/20`


src/jdk.javadoc/share/classes/jdk/javadoc/internal/doclets/formats/html/AnnotationTypeOptionalMemberWriterImpl.java
```diff
@@ -25,6 +25,7 @@
 
 package jdk.javadoc.internal.doclets.formats.html;
 
+import javax.lang.model.SourceVersion;
```
hns: Unused import

src/jdk.javadoc/share/classes/jdk/javadoc/internal/doclets/formats/html/HtmlConfiguration.java
```diff
@@ -181,7 +183,7 @@ public HtmlConfiguration(Doclet doclet, Locale locale, Reporter reporter) {
         }
 
         messages = new Messages(this, msgResources);
-        contents = new Contents(this);
+        //contents = new Contents(this);
```
hns: Should be removed instead of commented out

src/jdk.javadoc/share/classes/jdk/javadoc/internal/doclets/formats/html/HtmlConfiguration.java
```diff
@@ -195,6 +197,11 @@ public HtmlConfiguration(Doclet doclet, Locale locale, Reporter reporter) {
 
         conditionalPages = EnumSet.noneOf(ConditionalPage.class);
     }
+    protected void initConfiguration(DocletEnvironment docEnv,
+                                     Function<String, String> resourceKeyMapper) {
+        super.initConfiguration(docEnv, resourceKeyMapper);
+        contents = new Contents(this);
+    }
```
hns, line 203: I wish initialization of the configuration instance was not that complex... I guess the reason we must do it this way is that used SourceVersion is not known earlier.

test/langtools/jdk/javadoc/doclet/testTerminology/TestTerminology.java
```diff
@@ -0,0 +1,164 @@
...
+// awaits JDK-8247994: localize JavaScript search
+//        checkOutput("search.js", sv.compareTo(SourceVersion.RELEASE_16) < 0,
+//                """
+//                        var catTypes = "Types";""" //
+//        );
+//        checkOutput("search.js", sv.compareTo(SourceVersion.RELEASE_16) >= 0,
+//                """
+//                        var catTypes = "Classes and Interfaces";"""
+//        );
+// awaits JDK-8247994: localize JavaScript search
```
hns: Should be working now :)

hns (approved):
Looks good, Jon.

src/jdk.javadoc/share/classes/jdk/javadoc/internal/doclets/formats/html/HtmlConfiguration.java
jonathan-gibbons: Agreed that initializing the configuration is overly complex. At least in part, it's because of an earlier desire to use final fields where possible, when using direct access.   And yes, the problem here is that the `SourceVersion` is not known until `init` is called.

PR #21, opened 2020-12-15
8258134: assert(size == calc_size) failed: incorrect size calculation on x86_32 with AVX512 machines

DamonFool:
Hi all,

Two vector api tests crashed on x86_32 with AVX512 machines due to this assert [1].
The reason is that 'calc_size' is incorrect.

But there is no need to calculate 'calc_size' manually at all since the result [2] is actually never used by the VM.
Also, it is really hard to maintain the calculation logic for various hardwares and configurations.
And it may be easily broken again in the future with more and more complicated instructions & configurations.

So it would be better to remove the calculation and the assert, which is safe and already done for x86_64 [3].
The fix follows what is done for x86_64.
And it also makes vec_stack_to_stack_helper, vec_mov_helper and vec_spill_helper return void according to @neliasso 's comments [4].

Thanks.
Best regards,
Jie

[1] https://github.com/openjdk/jdk/blob/master/src/hotspot/cpu/x86/x86_32.ad#L1016
[2] https://github.com/openjdk/jdk/blob/master/src/hotspot/cpu/x86/x86_32.ad#L1059
[3] https://github.com/openjdk/jdk/blob/master/src/hotspot/cpu/x86/x86_64.ad#L1042
[4] https://github.com/openjdk/jdk/pull/1753#discussion_r542505836

---------
### Progress

### Issue
 * [JDK-8258134](https://bugs.openjdk.java.net/browse/JDK-8258134): assert(size == calc_size) failed: incorrect size calculation on x86_32 with AVX512 machines

### Reviewers

### Download
`$ git fetch https://git.openjdk.java.net/jdk16 pull/21/head:pull/21`
`$ git checkout pull/21`

### DIFF
--- a/src/hotspot/cpu/x86/gc/z/zBarrierSetAssembler_x86.cpp
+++ b/src/hotspot/cpu/x86/gc/z/zBarrierSetAssembler_x86.cpp
@@ -374,7 +374,7 @@ OptoReg::Name ZBarrierSetAssembler::refine_register(const Node* node, OptoReg::N
 }
 
 // We use the vec_spill_helper from the x86.ad file to avoid reinventing this wheel
-extern int vec_spill_helper(CodeBuffer *cbuf, bool do_size, bool is_load,
+extern int vec_spill_helper(CodeBuffer *cbuf, bool is_load,
                             int stack_offset, int reg, uint ireg, outputStream* st);
 
 #undef __
@@ -435,13 +435,13 @@ private:
     const OptoReg::Name opto_reg = OptoReg::as_OptoReg(reg_data._reg->as_VMReg());
     const uint ideal_reg = xmm_ideal_reg_for_size(reg_data._size);
     _spill_offset -= reg_data._size;
-    vec_spill_helper(__ code(), false /* do_size */, false /* is_load */, _spill_offset, opto_reg, ideal_reg, tty);
+    vec_spill_helper(__ code(), false /* is_load */, _spill_offset, opto_reg, ideal_reg, tty);
   }
 
   void xmm_register_restore(const XMMRegisterData& reg_data) {
     const OptoReg::Name opto_reg = OptoReg::as_OptoReg(reg_data._reg->as_VMReg());
     const uint ideal_reg = xmm_ideal_reg_for_size(reg_data._size);
-    vec_spill_helper(__ code(), false /* do_size */, true /* is_load */, _spill_offset, opto_reg, ideal_reg, tty);
+    vec_spill_helper(__ code(), true /* is_load */, _spill_offset, opto_reg, ideal_reg, tty);
     _spill_offset += reg_data._size;
   }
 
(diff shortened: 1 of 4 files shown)


DamonFool:
Testing:
  - tier1~tier3 on Linux/x86_{32, 64}, no regression

vnkozlov (approved):
Good.

DamonFool:
Thanks @vnkozlov for your review.

@neliasso , are you also OK with this change?
Thanks.

DamonFool:
May I get a second review for this fix?
Thanks.

TobiHartmann (approved):
Looks good.

DamonFool:
Thanks @TobiHartmann for your review.
/integrate

kevinrushforth:
This breaks the build for JDK 16 on Windows.

I see that the pre-submit tests were not run on this pull request. See the following information comment from the Skara bot in the [checks](https://github.com/openjdk/jdk16/pull/21/checks?check_run_id=1567324083) section for more information.

TobiHartmann:
> This breaks the build for JDK 16 on Windows.
> 
> I see that the pre-submit tests were not run on this pull request. See the following information comment from the Skara bot in the [checks](https://github.com/openjdk/jdk16/pull/21/checks?check_run_id=1567324083) section for more information.

I've filed [JDK-8258688](https://bugs.openjdk.java.net/browse/JDK-8258688).

TobiHartmann:
Okay, [JDK-8258687](https://bugs.openjdk.java.net/browse/JDK-8258687) has already been filed for this.

DamonFool:
Okay, I'll have a look. Sorry for this bug.

TobiHartmann:
The problem is that `extern int vec_spill_helper` has still `int` in the signature.

DamonFool:
> The problem is that `extern int vec_spill_helper` has still `int` in the signature.

Thanks @TobiHartmann for your help.
https://github.com/openjdk/jdk16/pull/51

I didn't know why the build tests were skipped for both pull/21 and pull/51.

kevinrushforth:
> I didn't know why the build tests were skipped for both pull/21 and pull/51.

You need to enable the GitHub actions to run on your repo. If you go to the checks link I added above, you will see:

_In order to run pre-submit tests, the source repository must be properly configured to allow test execution. See https://wiki.openjdk.java.net/display/SKARA/Testing for more information on how to configure this._

DamonFool:
> > I didn't know why the build tests were skipped for both pull/21 and pull/51.
> 
> You need to enable the GitHub actions to run on your repo. If you go to the checks link I added above, you will see:
> 
> _In order to run pre-submit tests, the source repository must be properly configured to allow test execution. See https://wiki.openjdk.java.net/display/SKARA/Testing for more information on how to configure this._

OK.
Thanks for your guidance.
I'll fix it next time.

PR #22, opened 2020-12-15
8255763: C2: OSR miscompilation caused by invalid memory instruction placement

robcasloz:
Disable GCM hoisting of memory-writing nodes for irreducible CFGs. This prevents GCM from wrongly "hoisting" stores into descendants of their original loop. Such an "inverted hoisting" can happen due to `CFGLoop::compute_freq()`'s inaccurate estimation of frequencies for irreducible CFGs.

Extend CFG verification code by checking that memory-writing nodes are placed in either their original loop or an ancestor.

Add tests for the reducible and irreducible cases. The former was already handled correctly before the change (the frequency estimation model prevents "inverted hoisting" for reducible CFGs), and is just added for coverage.

This change addresses the specific miscompilation issue in a conservative way, for simplicity and safety. Future work includes investigating if only the illegal blocks can be discarded as candidates for GCM hoisting, and refining frequency estimation for irreducible CFGs.

---------
### Progress

### Issue
 * [JDK-8255763](https://bugs.openjdk.java.net/browse/JDK-8255763): C2: OSR miscompilation caused by invalid memory instruction placement

### Reviewers
 * [Vladimir Kozlov](https://openjdk.java.net/census#kvn) (@vnkozlov - **Reviewer**) ⚠️ Review applies to 622299c3f2cc4934554f3ed4988418d2b16581d5

### Download
`$ git fetch https://git.openjdk.java.net/jdk16 pull/22/head:pull/22`
`$ git checkout pull/22`

### DIFF
--- a/src/hotspot/share/opto/block.cpp
+++ b/src/hotspot/share/opto/block.cpp
@@ -1226,7 +1226,7 @@ void PhaseCFG::verify() const {
       // of L. This is guaranteed by the freq. estimation model for reducible
       // CFGs, and by special handling in PhaseCFG::schedule_late() otherwise.
       if (n->is_Mach() && n->bottom_type()->has_memory() && n->in(0) != NULL) {
-        Block *original_block = find_block_for_node(n->in(0));
+        Block* original_block = find_block_for_node(n->in(0));
         assert(original_block != NULL, "missing block for memory-writing node");
         CFGLoop* original_or_ancestor = original_block->_loop;
         assert(block->_loop != NULL && original_or_ancestor != NULL, "no loop");
--- a/src/hotspot/share/opto/block.cpp
+++ b/src/hotspot/share/opto/block.cpp
@@ -1221,6 +1221,26 @@ void PhaseCFG::verify() const {
       if (j >= 1 && n->is_Mach() && n->as_Mach()->ideal_Opcode() == Op_CreateEx) {
         assert(j == 1 || block->get_node(j-1)->is_Phi(), "CreateEx must be first instruction in block");
       }
+      // Verify that memory-writing nodes (such as stores and calls) are placed
+      // in their original loop L (given by the control input) or in an ancestor
+      // of L. This is guaranteed by the freq. estimation model for reducible
+      // CFGs, and by special handling in PhaseCFG::schedule_late() otherwise.
+      if (n->is_Mach() && n->bottom_type()->has_memory() && n->in(0) != NULL) {
+        Block *original_block = find_block_for_node(n->in(0));
+        assert(original_block != NULL, "missing block for memory-writing node");
+        CFGLoop* original_or_ancestor = original_block->_loop;
+        assert(block->_loop != NULL && original_or_ancestor != NULL, "no loop");
+        bool found = false;
+        do {
+          if (block->_loop == original_or_ancestor) {
+            found = true;
+            break;
+          }
+          original_or_ancestor = original_or_ancestor->parent();
+        } while (original_or_ancestor != NULL);
+        assert(found, "memory-writing node is not placed in its original loop "
+                      "or an ancestor of it");
+      }
       if (n->needs_anti_dependence_check()) {
         verify_anti_dependences(block, n);
       }
--- a/src/hotspot/share/opto/block.hpp
+++ b/src/hotspot/share/opto/block.hpp
@@ -501,8 +501,8 @@ class PhaseCFG : public Phase {
   CFGLoop* create_loop_tree();
   bool is_dominator(Node* dom_node, Node* node);
   bool is_CFG(Node* n);
-  bool is_control_proj_or_safepoint(Node* n);
-  Block* find_block_for_node(Node* n);
+  bool is_control_proj_or_safepoint(Node* n) const;
+  Block* find_block_for_node(Node* n) const;
   bool is_dominating_control(Node* dom_ctrl, Node* n);
   #ifndef PRODUCT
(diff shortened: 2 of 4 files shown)


robcasloz:
/summary
Disable GCM hoisting of memory-writing nodes for irreducible CFGs. This prevents
GCM from wrongly "hoisting" stores into descendants of their original loop. Such
an "inverted hoisting" can happen due to CFGLoop::compute_freq()'s inaccurate
estimation of frequencies for irreducible CFGs.

Extend CFG verification code by checking that memory-writing nodes are placed in
either their original loop or an ancestor.

Add tests for the reducible and irreducible cases. The former was already
handled correctly before the change (the frequency estimation model prevents
"inverted hoisting" for reducible CFGs), and is just added for coverage.

This change addresses the specific miscompilation issue in a conservative way,
for simplicity and safety. Future work includes investigating if only the
illegal blocks can be discarded as candidates for GCM hoisting, and refining
frequency estimation for irreducible CFGs.

robcasloz:
Tested on `hs-tier1-9` on windows-x64, linux-x64, linux-aarch64, and macosx-x64 with `VerifyRegisterAllocator` enabled (to exercise all calls to the updated `PhaseCFG::verify()`).

robcasloz:
Tested for performance regressions on a set of standard benchmark suites (DaCapo, SPECjbb2005, SPECjvm2008, ...) and on windows-x64, linux-x64, and macosx-x64. No regression was observed, which can be expected since this change 1) only affects a minority of the compiled methods (those with irreducible CFGs), 2) only affects the placement of memory-writing nodes, which tends to be quite constrained already, and 3) forces the placement of these nodes, as much as possible, out of loops.

vnkozlov (commented):
Changes seems fine. And should be done regardless my following questions.

Is it possible to avoid "hoist inversion" if we take into account BCI and other information instead of only  frequency? Is it possible compute_freq() take into account irreducible loops?

robcasloz:
> Is it possible to avoid "hoist inversion" if we take into account BCI and other information instead of only frequency?

We could avoid it if

1) we followed the original heuristic proposed in Click's [Global code motion/global value numbering](https://courses.cs.washington.edu/courses/cse501/04wi/papers/click-pldi95.pdf) paper:

_"We choose the block that is in the shallowest loop nest possible, and then is as control dependent as possible."_

and

2) we had perfect loop information (including information about irreducible loops).

Unfortunately, information about irreducible loops is currently discarded when building the CFG-loop tree: 

https://github.com/openjdk/jdk16/blob/ce0ab2dd848484ad1837536942d42f19a509989b/src/hotspot/share/opto/gcm.cpp#L1657-L1659

> Is it possible compute_freq() take into account irreducible loops?

I think so. compute_freq() computes the frequencies of each block within a loop L in a single forward pass (in reverse DFS postorder) over the members (blocks and loops) of L:

https://github.com/openjdk/jdk16/blob/ce0ab2dd848484ad1837536942d42f19a509989b/src/hotspot/share/opto/gcm.cpp#L1790-L1808

In this computation, it assumes that the members form an acyclic graph (except for L's back-edges). This assumption does not hold for irreducible graphs, where there might be additional cycles corresponding to non-natural loops.  This could be refined by transforming the single forward pass into a proper system of equations to be solved iteratively. See some more details in the description field of the [bug report](https://bugs.openjdk.java.net/browse/JDK-8255763).

vnkozlov:
Thank you, @robcasloz
I propose to file RFE to do investigation about discussed improvements.
And proceed with current fix for JDK 16.

I don't see link to testing results in bug report.

robcasloz:
> I propose to file RFE to do investigation about discussed improvements.
> And proceed with current fix for JDK 16.

Thank you Vladimir for looking at this, I agree with your suggested plan.

vnkozlov (approved):
Testing results are good.

src/hotspot/share/opto/block.cpp
```diff
@@ -1221,6 +1221,26 @@ void PhaseCFG::verify() const {
       if (j >= 1 && n->is_Mach() && n->as_Mach()->ideal_Opcode() == Op_CreateEx) {
         assert(j == 1 || block->get_node(j-1)->is_Phi(), "CreateEx must be first instruction in block");
       }
+      // Verify that memory-writing nodes (such as stores and calls) are placed
+      // in their original loop L (given by the control input) or in an ancestor
+      // of L. This is guaranteed by the freq. estimation model for reducible
+      // CFGs, and by special handling in PhaseCFG::schedule_late() otherwise.
+      if (n->is_Mach() && n->bottom_type()->has_memory() && n->in(0) != NULL) {
+        Block *original_block = find_block_for_node(n->in(0));
```
chhagedorn: `*` should be at `Block`.

chhagedorn (approved):
Nice summary in the JBS issue! That looks good to me and I agree with Vladimir to do this fix in 16 and proceed with an RFE to further investigate the mentioned improvement possibilities.

src/hotspot/share/opto/block.cpp
robcasloz: Done, thanks!

robcasloz:
> Nice summary in the JBS issue! That looks good to me and I agree with Vladimir to do this fix in 16 and proceed with an RFE to further investigate the mentioned improvement possibilities.

Thanks for reviewing, Christian!

robcasloz:
Thanks for reviewing @vnkozlov and @chhagedorn! Requesting integration, since the only update after the reviews is a trivial style change (4f2763c).

chhagedorn (approved):

PR #24, opened 2020-12-15
8256843: [PPC64] runtime/logging/RedefineClasses.java fails with assert: registers not saved on stack

reinrich:
// Continuation of https://github.com/openjdk/jdk/pull/1724
// I'm taking over from @TheRealMDoerr who's busy with 11u downports

Please help review this small PPC64 fix.
It replaces the call os::current_stack_pointer() in os::current_frame()
with __builtin_frame_address(0) which is available also with xlC used by the AIX
build. With this os::current_frame() is agnostic to C++ compiler inlining.

Furthermore the fix changes os::current_frame() to return the correct
frame. Without the fix the returned frame is one frame to high which triggered
the assertion in trace_method_handle_stub().
This change uncovers an issue with NativeCallStack::NativeCallStack(int toSkip, bool fillStack)
on PPC64. There the call os::get_native_stack() is not compiled
to a tail call therefore the frame of the NativeCallStack constructor needs to
be skipped on PPC64 too as on other platforms.

Without this

make run-test TEST=test/hotspot/jtreg/runtime/NMT/CheckForProperDetailStackTrace.java

fails.

Furthermore the inline assembler from os::current_stack_pointer() is
removed. Instead __builtin_frame_address(0) is used again.

The fix passed our CI testing: JCK and JTREG, also in Xcomp mode, SPECjvm2008, SPECjbb2015, Renaissance Suite,
SAP specific tests with fastdebug and release builds on all platforms

---------
### Progress

### Issue
 * [JDK-8256843](https://bugs.openjdk.java.net/browse/JDK-8256843): [PPC64] runtime/logging/RedefineClasses.java fails with assert: registers not saved on stack

### Reviewers
 * [Martin Doerr](https://openjdk.java.net/census#mdoerr) (@TheRealMDoerr - **Reviewer**)
 * [Lutz Schmidt](https://openjdk.java.net/census#lucy) (@RealLucy - **Reviewer**)

### Download
`$ git fetch https://git.openjdk.java.net/jdk16 pull/24/head:pull/24`
`$ git checkout pull/24`

### DIFF
--- a/src/hotspot/os_cpu/aix_ppc/os_aix_ppc.cpp
+++ b/src/hotspot/os_cpu/aix_ppc/os_aix_ppc.cpp
@@ -65,16 +65,7 @@
 # include <ucontext.h>
 
 address os::current_stack_pointer() {
-  address csp;
-
-#if !defined(USE_XLC_BUILTINS)
-  // inline assembly for `mr regno(csp), R1_SP':
-  __asm__ __volatile__ ("mr %0, 1":"=r"(csp):);
-#else
-  csp = (address) __builtin_frame_address(0);
-#endif
-
-  return csp;
+  return (address)__builtin_frame_address(0);
 }
 
 char* os::non_memory_address_word() {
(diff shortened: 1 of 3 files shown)


TheRealMDoerr (approved):
Thank you for fixing it in jdk16 as discussed in the jdk PR!
Note that PR and JBS title must match, otherwise integration is blocked.

reinrich:
> Thank you for fixing it in jdk16 as discussed in the jdk PR!

Thanks for the review!

> Note that PR and JBS title must match, otherwise integration is blocked.

Mon dieu! I corrected the PR title :)

RealLucy (approved):
Changes look good. Thanks for fixing.

Side note (not relevant for approval): Wouldn't it be nice to use the builtin just once and call os::current_stack_pointer() in os::current_frame()?

reinrich:
> 
> 
> Changes look good. Thanks for fixing.

Thanks for the review!

> Side note (not relevant for approval): Wouldn't it be nice to use the builtin just once and call os::current_stack_pointer() in os::current_frame()?

Yeah this would be nice but then we would be dependent to call inlining by the C++ compiler, i.e. we'd have to fiddle around with _NMT_NOINLINE_ again (btw: we could do the same on [s390 where _NMT_NOINLINE_ is used](https://github.com/openjdk/jdk/blob/83be8a902cee867c0e9d400e762f61896eb6df80/src/hotspot/os_cpu/linux_s390/os_linux_s390.cpp#L197-L203))

reinrich:
Thanks again for the reviews @TheRealMDoerr, @RealLucy!

/integrate

PR #30, opened 2020-12-15
8258236: Segfault in ClassListParser::resolve_indy dumping static AppCDS archive

calvinccheung:
Please review this change for JDK 16.

In ClassListParser::resolve_indy, if a class has previously failed verification, don't proceed with resolve indy for that class to avoid dereferencing a null cpcache pointer.

Passed tiers 1,2,3,4 tests.

---------
### Progress

### Issue
 * [JDK-8258236](https://bugs.openjdk.java.net/browse/JDK-8258236): Segfault in ClassListParser::resolve_indy dumping static AppCDS archive

### Reviewers
 * [Coleen Phillimore](https://openjdk.java.net/census#coleenp) (@coleenp - **Reviewer**)

### Download
`$ git fetch https://git.openjdk.java.net/jdk16 pull/30/head:pull/30`
`$ git checkout pull/30`

### DIFF
--- a/src/hotspot/share/classfile/classListParser.cpp
+++ b/src/hotspot/share/classfile/classListParser.cpp
@@ -467,6 +467,10 @@ void ClassListParser::resolve_indy(Symbol* class_name_symbol, TRAPS) {
   Klass* klass = SystemDictionary::resolve_or_fail(class_name_symbol, class_loader, protection_domain, true, THREAD); // FIXME should really be just a lookup
   if (klass != NULL && klass->is_instance_klass()) {
     InstanceKlass* ik = InstanceKlass::cast(klass);
+    if (SystemDictionaryShared::has_class_failed_verification(ik)) {
+      // don't attempt to resolve indy on classes that has previously failed verification
+      return;
+    }
     MetaspaceShared::try_link_class(ik, THREAD);
     assert(!HAS_PENDING_EXCEPTION, "unexpected exception");
 
--- a/test/hotspot/jtreg/TEST.groups
+++ b/test/hotspot/jtreg/TEST.groups
@@ -336,6 +336,7 @@ hotspot_appcds_dynamic = \
  -runtime/cds/appcds/ExtraSymbols.java \
  -runtime/cds/appcds/LambdaEagerInit.java \
  -runtime/cds/appcds/LambdaProxyClasslist.java \
+ -runtime/cds/appcds/LambdaVerificationFailedDuringDump.java \
  -runtime/cds/appcds/LongClassListPath.java \
  -runtime/cds/appcds/LotsOfClasses.java \
  -runtime/cds/appcds/NonExistClasspath.java \
--- a/test/hotspot/jtreg/runtime/cds/appcds/LambdaVerificationFailedDuringDump.java
+++ b/test/hotspot/jtreg/runtime/cds/appcds/LambdaVerificationFailedDuringDump.java
@@ -1,0 +1,50 @@
+/*
+ * Copyright (c) 2020, Oracle and/or its affiliates. All rights reserved.
+ * DO NOT ALTER OR REMOVE COPYRIGHT NOTICES OR THIS FILE HEADER.
+ *
+ * This code is free software; you can redistribute it and/or modify it
+ * under the terms of the GNU General Public License version 2 only, as
+ * published by the Free Software Foundation.
+ *
+ * This code is distributed in the hope that it will be useful, but WITHOUT
+ * ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or
+ * FITNESS FOR A PARTICULAR PURPOSE.  See the GNU General Public License
+ * version 2 for more details (a copy is included in the LICENSE file that
+ * accompanied this code).
+ *
+ * You should have received a copy of the GNU General Public License version
+ * 2 along with this work; if not, write to the Free Software Foundation,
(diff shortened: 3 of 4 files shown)


cl4es (approved):
Fix and tests looks good!

src/hotspot/share/classfile/classListParser.cpp
```diff
@@ -467,6 +467,10 @@ void ClassListParser::resolve_indy(Symbol* class_name_symbol, TRAPS) {
   Klass* klass = SystemDictionary::resolve_or_fail(class_name_symbol, class_loader, protection_domain, true, THREAD); // FIXME should really be just a lookup
   if (klass != NULL && klass->is_instance_klass()) {
     InstanceKlass* ik = InstanceKlass::cast(klass);
+    if (SystemDictionaryShared::has_class_failed_verification(ik)) {
+      // don't attempt to resolve indy on classes that has previously failed verification
+      return;
+    }
     MetaspaceShared::try_link_class(ik, THREAD);
```
coleenp, line 474: Doesn't the check for failing verification belong after try_link_class(), which runs the verifier?

coleenp (changes requested):

src/hotspot/share/classfile/classListParser.cpp
```diff
@@ -467,6 +467,10 @@ void ClassListParser::resolve_indy(Symbol* class_name_symbol, TRAPS) {
   Klass* klass = SystemDictionary::resolve_or_fail(class_name_symbol, class_loader, protection_domain, true, THREAD); // FIXME should really be just a lookup
```
coleenp, line 467: If an exception is unexpected, this should be CHECK not THREAD.

src/hotspot/share/classfile/classListParser.cpp
calvinccheung: Linking attempt was done for the class and was marked as failed verification prior to getting into `ClassListParser::resolve_indy`.

Consider the following class list (from the test case):
```
BadInvokeDynamic
@lambda-proxy BadInvokeDynamic run ()Ljava/lang/Runnable; ()V REF_invokeStatic BadInvokeDynamic lambda$doTest$0 ()V ()V
```

The BadInvokeDynamic was linked via the following code path:
```
	MetaspaceShared::try_link_class() at metaspaceShared.cpp:1,161 0x7ffff63b1c10	
	MetaspaceShared::preload_classes() at metaspaceShared.cpp:1,125 0x7ffff63b1a42	
	MetaspaceShared::preload_and_dump() at metaspaceShared.cpp:1,047 0x7ffff63b1557	
	Threads::create_vm() at thread.cpp:3,809 0x7ffff666b3f9	
	JNI_CreateJavaVM_inner() at jni.cpp:3,769 0x7ffff604b1be	
	JNI_CreateJavaVM() at jni.cpp:3,852 0x7ffff604b4c3	
	InitializeJVM() at java.c:1,536 0x7ffff7fcd664	
	JavaMain() at java.c:416 0x7ffff7fca252	
	ThreadJavaMain() at java_md.c:651 0x7ffff7fd0f77	
	start_thread() at 0x7ffff79b0ea5
```

It was marked as failed verification in `MetaspaceShared::try_link_class` in the following:
```
    ik->link_class(THREAD);
    if (HAS_PENDING_EXCEPTION) {
      ResourceMark rm(THREAD);
      log_warning(cds)("Preload Warning: Verification failed for %s",
                    ik->external_name());
      CLEAR_PENDING_EXCEPTION;
      SystemDictionaryShared::set_class_has_failed_verification(ik);
      _has_error_classes = true;
    }
```
During the processing of the `@lambda-proxy` in the class list, the `ClassListParser::resolve_indy` would be called and the check is for checking if the caller class has failed verification before.

src/hotspot/share/classfile/classListParser.cpp
calvinccheung: I'll try CHECK and do more testing before posting an updated webrev.

Thanks for your review.

calvinccheung:
> Fix and tests looks good!

@cl4es Thanks for your review.

src/hotspot/share/classfile/classListParser.cpp
coleenp: For this change, don't change it to CHECK in case something new will fail, and any new exceptions will caught by checking k != NULL.  It just looks strange but can be cleaned up later.

src/hotspot/share/classfile/classListParser.cpp
coleenp: Okay so you're trying to catch a previous link time failure.  Why do you expect the next try_link_class call to never fail?

coleenp (approved):
I just have questions.  The change as-is is good.

calvinccheung:
>For this change, don't change it to CHECK in case something new will fail, and any new exceptions will caught by checking k != NULL. It just looks strange but can be cleaned up later.

I was thinking the same - do the clean up later. Initial local testing looks good - no new cds/appcds tests failure.

>Okay so you're trying to catch a previous link time failure. Why do you expect the next try_link_class call to never fail?

Previous link time failure was recorded via `SystemDictionaryShared::set_class_has_failed_verification(ik)`.
`try_link_class` will not result in an exception because there's a `if (HAS_PENDING_EXCEPTION)` block which calls `CLEAR_PENDING_EXCEPTION`.

PR #33, opened 2020-12-16
8256693: getAnnotatedReceiverType parameterizes types too eagerly

jbf:
The fix for JDK-8256693 too often produces a ParameterizedType as the result of getAnnotatedReceiverType().getType() . A ParameterizedType is necessary only when this type or any of its transitive owner types has type parameters, but should be avoided if this isn't the case.

This implementation recursively creates a chain of ParameterizedTypes starting from the outermost type that has type parameters.

See here for the now closed JDK 17 pr: https://github.com/openjdk/jdk/pull/1414

---------
### Progress

### Issue
 * [JDK-8256693](https://bugs.openjdk.java.net/browse/JDK-8256693): getAnnotatedReceiverType parameterizes types too eagerly

### Reviewers
 * [Vicente Romero](https://openjdk.java.net/census#vromero) (@vicente-romero-oracle - **Reviewer**)

### Download
`$ git fetch https://git.openjdk.java.net/jdk16 pull/33/head:pull/33`
`$ git checkout pull/33`

### DIFF
--- a/test/jdk/java/lang/annotation/typeAnnotations/GetAnnotatedReceiverType.java
+++ b/test/jdk/java/lang/annotation/typeAnnotations/GetAnnotatedReceiverType.java
@@ -26,10 +26,7 @@
  * @bug 8024915 8044629 8256693
  */
 
-import java.lang.reflect.AnnotatedType;
-import java.lang.reflect.Executable;
-import java.lang.reflect.ParameterizedType;
-import java.lang.reflect.Type;
+import java.lang.reflect.*;
 
 public class GetAnnotatedReceiverType {
     public void method() {}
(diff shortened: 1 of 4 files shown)


test/jdk/java/lang/annotation/typeAnnotations/GetAnnotatedReceiverType.java
```diff
@@ -132,9 +150,35 @@ public static void main(String[] args) throws NoSuchMethodException {
...
+
+        Inner2.Inner3.Inner7<?> instance7 = instance3.new Inner7<String>();
+        checkAnnotatedReceiverType(instance7.getClass().getMethod("innerMethod7"), true,
+                "The type of .getAnnotatedReceiverType().getType() for this method should be");
+        recursiveCheckAnnotatedOwnerTypes(instance7.getClass().getMethod("innerMethod7").getAnnotatedReceiverType());
+
         if (failures != 0)
             throw new RuntimeException("Test failed, see log for details");
-        else if (tests != 15)
+        else if (tests != 25)
```
vicente-romero-oracle: this looks a bit brittle, isn't it better to count the number of tests failing and issue an error if that number is > 0?

vicente-romero-oracle (changes requested):

test/jdk/java/lang/annotation/typeAnnotations/GetAnnotatedReceiverType.java
jbf: Yes, not great, but at least it isn't brittle when running the test, only when changing it. I'd like to take a separate pass over the tests for 17 if possible.

vicente-romero-oracle: > Yes, not great, but at least it isn't brittle when running the test, only when changing it. I'd like to take a separate pass over the tests for 17 if possible.

what about declaring a static final field for that value instead of using a literal?

jbf: Done

vicente-romero-oracle (approved):

PR #38, opened 2020-12-16
8255880: UI of Swing components is not redrawn after their internal state changed

alitvinv:
Hello,

This is a review request for straight back port of the fix for the bug JDK-8255880 from JDK 17 to JDK 16, which is currently in RDP 1 phase. The fix was tested and the regression test was executed on MS Windows OS, Linux OS, macOS.

Thank you,
Anton

---------
### Progress

### Issue
 * [JDK-8255880](https://bugs.openjdk.java.net/browse/JDK-8255880): UI of Swing components is not redrawn after their internal state changed

### Reviewers
 * [Phil Race](https://openjdk.java.net/census#prr) (@prrace - **Reviewer**)

### Download
`$ git fetch https://git.openjdk.java.net/jdk16 pull/38/head:pull/38`
`$ git checkout pull/38`

### DIFF
--- a/src/java.desktop/unix/classes/sun/awt/X11/XFramePeer.java
+++ b/src/java.desktop/unix/classes/sun/awt/X11/XFramePeer.java
@@ -341,6 +341,13 @@ class XFramePeer extends XDecoratedPeer implements FramePeer {
             }
         }
         handleStateChange(old_state, state);
+
+        // RepaintManager does not repaint iconified windows. Window needs to be
+        // repainted explicitly, when it is deiconified.
+        if (((changed & Frame.ICONIFIED) != 0) &&
+            ((state & Frame.ICONIFIED) == 0)) {
+            repaint();
+        }
     }
 
     // NOTE: This method may be called by privileged threads.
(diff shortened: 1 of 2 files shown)


prrace (approved):
Fine but in that casw why didn't you withdraw the 17 PR and just push to 16 which will get forward synced ?

alitvinv:
Hello Phil. For me the main goal was to fix this bug in the latest JDK 
feature release, that is why I initiated integration to "openjdk/jdk" as 
soon as the fix was approved there. Integration of this fix to 
"openjdk/jdk" repository was my first integration to OpenJDK project in 
Git, at that moment I did not anticipate that obligatory synchronization 
from "openjdk/jdk16" to "openjdk/jdk" yet. Today based on the fact that 
this bug is P3 and not an enhancement and that following "fix-request" 
process in JBS is not prescribed for this type of bug in RDP 1 phase, 
and that this bug on Monday still was targeted to JDK 16, I came to a 
conclusion that there is legal right to go through this back port 
procedure to try to deliver the fix to JDK 16. I thought that there 
would not be a trouble during this synchronization, because in Mercurial 
such cases during merge were handled automatically silently. Hopefully 
Git will manage to handle it automatically.

I guarantee that this is the first and last time from my side, next 
times, if there are, I will not be initiating back ports to 
stabilization repositories in pre-RDP 2 phases, if the fix is integrated 
to always open repository.

PR #41, opened 2020-12-17
8258393: Shenandoah: "graph should be schedulable" assert failure

rwestrel:
This is a shenandoah bug but the fix is in shared code.

At barrier expansion time, we need the raw memory state for the
control at which the barrier is expanded. Before expansion, the memory
graph is traversed and memory state for each control is recorded. In
that process, if opportunities to improve the memory graph are found,
the graph is modified. In the case of the failure, a raw memory load
was assigned a call projection as control but for that control, no
memory state was recorded and the fall back is to use the call's input
memory state. As a consequence the load's memory input is updated to a
wrong memory state. This causes the assert failure.

The call has 2 memory projections: one for the fallthrough and one for
the exception path. One projection is the memory state for the
CatchProj for the fallthrough, the other one for the CatchProj of the
exception path. For the control projection out of the call, there's no
memory state that makes sense and assigning the control projection of
the call as control for the load is the root cause of the
problem. This happens because anti-dependence analysis is too
conservative: a memory Phi that merges states from the exception and
fallthrough path is considered a dependency and that pushed the load
right after the call.

I propose to be less conservative in anti-dependence analysis for
Phis. For a Phi, when computing the LCA, I think it's sufficient to
only consider region's inputs that we actually reach by following the
memory edges and that's what I propose here.

---------
### Progress

### Issue
 * [JDK-8258393](https://bugs.openjdk.java.net/browse/JDK-8258393): Shenandoah: "graph should be schedulable" assert failure

### Reviewers
 * [Roman Kennke](https://openjdk.java.net/census#rkennke) (@rkennke - **Reviewer**) ⚠️ Review applies to 31f3bd1c9b01c3e9cf4a071eca7885dcefa01fbe

### Download
`$ git fetch https://git.openjdk.java.net/jdk16 pull/41/head:pull/41`
`$ git checkout pull/41`

### DIFF
--- a/src/hotspot/share/gc/shenandoah/c2/shenandoahSupport.cpp
+++ b/src/hotspot/share/gc/shenandoah/c2/shenandoahSupport.cpp
@@ -2459,6 +2459,21 @@ bool MemoryGraphFixer::mem_is_valid(Node* m, Node* c) const {
 
 Node* MemoryGraphFixer::find_mem(Node* ctrl, Node* n) const {
   assert(n == NULL || _phase->ctrl_or_self(n) == ctrl, "");
+  assert(!ctrl->is_Call() || ctrl == n, "projection expected");
+#ifdef ASSERT
+  if ((ctrl->is_Proj() && ctrl->in(0)->is_Call()) ||
+          (ctrl->is_Catch() && ctrl->in(0)->in(0)->is_Call())){
+    CallNode* call = ctrl->is_Proj() ? ctrl->in(0)->as_Call() : ctrl->in(0)->in(0)->as_Call();
+    int mems = 0;
+    for (DUIterator_Fast imax, i = call->fast_outs(imax); i < imax; i++) {
+      Node* u = call->fast_out(i);
+      if (u->bottom_type() == Type::MEMORY) {
+        mems++;
+      }
+    }
+    assert(mems <= 1, "No node right after call if multiple mem projections");
+  }
+#endif
   Node* mem = _memory_nodes[ctrl->_idx];
   Node* c = ctrl;
   while (!mem_is_valid(mem, c) &&
--- a/src/hotspot/share/gc/shenandoah/c2/shenandoahSupport.cpp
+++ b/src/hotspot/share/gc/shenandoah/c2/shenandoahSupport.cpp
@@ -2462,7 +2462,7 @@ Node* MemoryGraphFixer::find_mem(Node* ctrl, Node* n) const {
   assert(!ctrl->is_Call() || ctrl == n, "projection expected");
 #ifdef ASSERT
   if ((ctrl->is_Proj() && ctrl->in(0)->is_Call()) ||
-          (ctrl->is_Catch() && ctrl->in(0)->in(0)->is_Call())){
+      (ctrl->is_Catch() && ctrl->in(0)->in(0)->is_Call())) {
     CallNode* call = ctrl->is_Proj() ? ctrl->in(0)->as_Call() : ctrl->in(0)->in(0)->as_Call();
     int mems = 0;
     for (DUIterator_Fast imax, i = call->fast_outs(imax); i < imax; i++) {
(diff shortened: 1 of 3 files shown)


rkennke (approved):
As far as I can tell, the change looks good. I'm not very familiar with the code, it would be good to have somebody else check it too.
I verified that the tests that failed before are passing now and also have run some additional tests.

test/hotspot/jtreg/gc/shenandoah/compiler/TestBadRawMemoryAfterCall.java
```diff
@@ -0,0 +1,70 @@
...
+ * Inc., 51 Franklin St, Fifth Floor, Boston, MA 02110-1301 USA.
+ *
+ * Please contact Oracle, 500 Oracle Parkway, Redwood Shores, CA 94065 USA
+ * or visit www.oracle.com if you need additional information or have any
+ * questions.
+ */
+
+/**
+ * @test
+ *
```
TobiHartmann, line 30: Missing `@requires vm.gc.Shenandoah`

test/hotspot/jtreg/gc/shenandoah/compiler/TestBadRawMemoryAfterCall.java
```diff
@@ -0,0 +1,70 @@
+/*
+ * Copyright (c) 2020, Red Hat, Inc. All rights reserved.
```
TobiHartmann: Copyright should be updated to 2021.

src/hotspot/share/gc/shenandoah/c2/shenandoahSupport.cpp
```diff
@@ -2459,6 +2459,21 @@ bool MemoryGraphFixer::mem_is_valid(Node* m, Node* c) const {
 
 Node* MemoryGraphFixer::find_mem(Node* ctrl, Node* n) const {
   assert(n == NULL || _phase->ctrl_or_self(n) == ctrl, "");
+  assert(!ctrl->is_Call() || ctrl == n, "projection expected");
+#ifdef ASSERT
+  if ((ctrl->is_Proj() && ctrl->in(0)->is_Call()) ||
+          (ctrl->is_Catch() && ctrl->in(0)->in(0)->is_Call())){
```
TobiHartmann: Indentation is wrong and whitespace is missing here `))){`

src/hotspot/share/opto/loopnode.cpp
```diff
@@ -5011,6 +5011,22 @@ Node *PhaseIdealLoop::get_late_ctrl( Node *n, Node *early ) {
...
+      // For Phis only consider Region's inputs that were reached by following the memory edges
+      if (LCA != early) {
+        for (uint i = 0; i < worklist.size(); i++) {
+          Node* s = worklist.at(i);
+          if (s->is_Phi() && C->can_alias(s->adr_type(), load_alias_idx)) {
+            Node* r = s->in(0);
+            for (uint j = 1; j < s->req(); j++) {
+              Node* in = s->in(j);
+              Node* r_in = r->in(j);
+              if (worklist.member(in) && is_dominator(early, r_in)) {
```
TobiHartmann, line 5023: Above we check for `!sctrl->is_top()`, do we need the same check for `r_in`?

TobiHartmann (approved):
Otherwise looks reasonable to me.

src/hotspot/share/opto/loopnode.cpp
rwestrel: I don't think a Region can have a top input at this point (previous IGVN should have taken care of it).

rwestrel:
> Otherwise looks reasonable to me.

Thanks for the review. I took care of your comments (except for the is_top() check that I don't think is needed).

TobiHartmann (approved):
Thanks for updating, looks good to me.

rwestrel:
> As far as I can tell, the change looks good. I'm not very familiar with the code, it would be good to have somebody else check it too.
> I verified that the tests that failed before are passing now and also have run some additional tests.

thanks for the review and running tests.

PR #42, opened 2020-12-17
8257974: Regression 21% in DaCapo-lusearch-large after JDK-8236926

kstefanj:
Please review this fix to avoid the regression in DaCapo-lusearch-large.

**Summary** 
Doing uncommit concurrently with this benchmark was the cause of this regression. Using JFR we can see a lot of the java threads are blocking on a monitor when using the build with concurrent uncommit (but not otherwise). We know that the uncommit will affect the application threads and if a thread holding a lock gets stalled this will affect the overall performance more than expected. To avoid stalling the application threads more than necessary we can be a bit less aggressive when doing the uncommit. The initial version of concurrent uncommit tries to return the memory as quickly as possible after the GC, but there is no contract that we need to return the memory that quick and doing it a bit more lazy have more than one positive effect. 

The proposed change does three things: 
* first we delay the first uncommit after the GC shrinking the heap by 100ms, 
* after that each new invocation of the uncommit task will be delayed by 10ms (instead of run back to back)
* the size of each uncommit is changed from 256m to 128m, this will also lower the impact of each uncommit call. 

Doing these things will make the uncommit take a longer time and for applications where the uncommitted memory is needed quite quickly again, this will have the nice effect that instead of uncommitting/committing, we can just reuse the memory without involving the OS.

We have also done some experiments using `madvise` to do the uncommitting, but this doesn't fix the whole regression and needs some more investigation. A bit more information around this can be found in the bug report.

**Testing**
Tier1-3 in Mach5 and a lot of manual performance testing both locally and in our internal systems. Regression has been removed and in we instead see a small improvement, due to being able to reuse memory instead of uncommiting and then committing it again.

---------
### Progress

### Issue
 * [JDK-8257974](https://bugs.openjdk.java.net/browse/JDK-8257974): Regression 21% in DaCapo-lusearch-large after JDK-8236926

### Reviewers
 * [Albert Mingkun Yang](https://openjdk.java.net/census#ayang) (@albertnetymk - Author) ⚠️ Review applies to 2d5c88b466bcad900b4555e27da0396aac6bdbe3
 * [Claes Redestad](https://openjdk.java.net/census#redestad) (@cl4es - **Reviewer**) ⚠️ Review applies to 2d5c88b466bcad900b4555e27da0396aac6bdbe3
 * [Thomas Schatzl](https://openjdk.java.net/census#tschatzl) (@tschatzl - **Reviewer**) ⚠️ Review applies to 2d5c88b466bcad900b4555e27da0396aac6bdbe3

### Download
`$ git fetch https://git.openjdk.java.net/jdk16 pull/42/head:pull/42`
`$ git checkout pull/42`

### DIFF
--- a/src/hotspot/share/gc/g1/g1UncommitRegionTask.cpp
+++ b/src/hotspot/share/gc/g1/g1UncommitRegionTask.cpp
@@ -59,7 +59,7 @@ void G1UncommitRegionTask::enqueue() {
 
   G1UncommitRegionTask* uncommit_task = instance();
   if (!uncommit_task->is_active()) {
-    // Change state to active and schedule with no delay.
+    // Change state to active and schedule using UncommitInitialDelayMs.
     uncommit_task->set_active(true);
     G1CollectedHeap::heap()->service_thread()->schedule_task(uncommit_task, UncommitInitialDelayMs);
   }
--- a/src/hotspot/share/gc/g1/g1UncommitRegionTask.cpp
+++ b/src/hotspot/share/gc/g1/g1UncommitRegionTask.cpp
@@ -61,7 +61,7 @@ void G1UncommitRegionTask::enqueue() {
   if (!uncommit_task->is_active()) {
     // Change state to active and schedule with no delay.
     uncommit_task->set_active(true);
-    G1CollectedHeap::heap()->service_thread()->schedule_task(uncommit_task, 0);
+    G1CollectedHeap::heap()->service_thread()->schedule_task(uncommit_task, UncommitInitialDelayMs);
   }
 }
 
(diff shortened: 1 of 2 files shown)


src/hotspot/share/gc/g1/g1UncommitRegionTask.cpp
```diff
@@ -61,7 +61,7 @@ void G1UncommitRegionTask::enqueue() {
   if (!uncommit_task->is_active()) {
     // Change state to active and schedule with no delay.
```
cl4es: Comment needs an update

albertnetymk (approved):

cl4es (approved):
LGTM - seems like a reasonable approach. I'm sure an argument can be made for being able to control the delays (I can picture situations where tuning in either direction might make sense), but I agree with fixing this regression by making these delays constant for now.

tschatzl (approved):
Lgtm sans the comment update.

src/hotspot/share/gc/g1/g1UncommitRegionTask.cpp
kstefanj: Nice catch!

kstefanj:
> LGTM - seems like a reasonable approach. I'm sure an argument can be made for being able to control the delays (I can picture situations where tuning in either direction might make sense), but I agree with fixing this regression by making these delays constant for now.

Thanks for the review.

Yes, we might end up adding flags for these delays in a later release. Or we might tune the values for the constants if we find something that seem to work better. As long as we don't see this causing big problems, I would prefer if we can have well balanced constants, instead of adding flags that don't add much value.

kstefanj:
Thanks @cl4es, @albertnetymk and @tschatzl for the reviews.

/integrate

PR #44, opened 2020-12-17
8258225: compiler/c2/cr6340864/TestIntVect.java runs faster in interpreter

neliasso:
Exclude test-method from compiler/c2/cr6340864-tests.

In debug builds - the C1 method IR::verify is very costly in big methods with many blocks. The compile times explode. A sub minute test takes 35 minutes on slower machines - and these tests are run multiple times.

Some of these tests are saved by the HugeMethodLimit, but some (TestIntVect and TestLongVect) doesn't.

Please review,
Nils

---------
### Progress

### Issue
 * [JDK-8258225](https://bugs.openjdk.java.net/browse/JDK-8258225): compiler/c2/cr6340864/TestIntVect.java runs faster in interpreter

### Reviewers
 * [Vladimir Ivanov](https://openjdk.java.net/census#vlivanov) (@iwanowww - **Reviewer**) ⚠️ Review applies to 255b456c588088030189f369f85c0a550f7d2632

### Download
`$ git fetch https://git.openjdk.java.net/jdk16 pull/44/head:pull/44`
`$ git checkout pull/44`

### DIFF
--- a/test/hotspot/jtreg/compiler/c2/cr6340864/TestByteVect.java
+++ b/test/hotspot/jtreg/compiler/c2/cr6340864/TestByteVect.java
@@ -26,10 +26,10 @@
  * @bug 6340864
  * @summary Implement vectorization optimizations in hotspot-server
  *
- * @run main/othervm/timeout=400 -Xbatch -XX:CompileCommand=exclude,*::test() -Xmx128m compiler.c2.cr6340864.TestByteVect
- * @run main/othervm/timeout=400 -Xbatch -XX:CompileCommand=exclude,*::test() -Xmx128m -XX:MaxVectorSize=8 compiler.c2.cr6340864.TestByteVect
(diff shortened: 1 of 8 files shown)


iwanowww (approved):
Looks good.

(Wasn't aware about `class::method()` syntax support in -XX:CompileCommand.)

Does it make sense to reconsider the timeout value (from 400 to a lower value)?

TobiHartmann (approved):
Looks good.

neliasso:
Thank you Vladimir and Tobias!

> Does it make sense to reconsider the timeout value (from 400 to a lower value)?

Yes it a good suggestion. I will remove the timout so that the default of 120 (before scaling) is used.

TobiHartmann (approved):

vnkozlov (approved):

PR #45, opened 2020-12-17
8258619: -XX:CompileCommand=blackhole,* should only accept empty static methods

shipilev:
In JDK-8257827 CSR, a further comment was raised about changing the semantics of the method. This allegedly introduces the conformance problem. While it is still an open question where the conformance problem is real, we can sidestep the entire issue by allowing only empty static methods to be blackholed. This eliminates any behavioral difference: empty methods have no side effects, and static methods do not need any null checks.

This change eliminates a lot of new code added by original JDK-8252505 push, mostly now-useless testing code. It adds two more tests that verify that non-empty and non-static methods are properly rejected.

---------
### Progress

### Issue
 * [JDK-8258619](https://bugs.openjdk.java.net/browse/JDK-8258619): -XX:CompileCommand=blackhole,* should only accept empty static methods

### Reviewers

### Download
`$ git fetch https://git.openjdk.java.net/jdk16 pull/45/head:pull/45`
`$ git checkout pull/45`


TobiHartmann (approved):
Copyrights should be updated to 2021. Otherwise looks good to me.

chhagedorn (approved):
Looks good to me!

vnkozlov (approved):
I agree with these changes for JDK 16.

PR #49, opened 2020-12-17
8213032: program fails with LambdaConversionException at execution time

vicente-romero-oracle:
Hi,

Currently javac is accepting this code:

```
import java.util.stream.*;

public class Test {
    interface I {}
    static abstract class C { }
    static class A extends C implements I { }
    static class B extends C implements I { }

    static String f(I i) { return null; }

    public static void main(String[] args) {
        Stream.of(new A(), new B()).map(Test::f).forEach(System.out::println);
    }
}
```

the line with the method reference can also be written using a lambda expression like in:

```
        Stream.of(new A(), new B()).map(i -> f(i)).forEach(System.out::println);
```
but the code with the method reference, is failing later at execution time with `LambdaConversionException`. Currently javac is desugaring the method reference to the equivalent lambda but the argument of generated lambda is: `Test$I` at the same time the target type of the method reference has argument with type `C & I` which will be desugared by the compiler to `C`. This is the reason for the runtime error. The right type for the argument of the desugared method generated by javac is `C` which is what this patch is doing. The change to method `ReferenceTranslationContext::bridgedRefSig` has been done in order to obtain a more exact version of the target type, as the current implementation is returning `Object` in most cases, which is OK, but I think that we could generate a better, more accurate, type.

TIA for the review

---------
### Progress

### Issue
 * [JDK-8213032](https://bugs.openjdk.java.net/browse/JDK-8213032): program fails with LambdaConversionException at execution time

### Reviewers
 * [Jan Lahoda](https://openjdk.java.net/census#jlahoda) (@lahodaj - **Reviewer**) ⚠️ Review applies to 4d8ef5d826b4ab6e240de161ab4473174c46edff

### Download
`$ git fetch https://git.openjdk.java.net/jdk16 pull/49/head:pull/49`
`$ git checkout pull/49`


lahodaj (approved):
Looks reasonable.

PR #50, opened 2020-12-18
8258384: AArch64: SVE verify_ptrue fails on some tests

nsjian:
After applying [1], some Vector API tests fail with SIGILL on SVE
system. The SIGILL was triggered by verify_ptrue before c2 compiled
function returns, which means that the preserved p7 register (as ptrue)
has been clobbered before returning to c2 compiled code. (p7 is not
preserved cross function calls, and system calls [2]).

Currently we try to reinitialize ptrue at each entrypoint of returning
from non-c2 compiled code, which indicating possible C or system calls.
However, there's still one entrypoint missing, exception handling, as
we may jump to c2 compiled code for exception handler. See
OptoRuntime::generate_exception_blob().

Adding reinitialize_ptrue before jumping back to c2 compiled code in
generate_exception_blob() could solve those Vector API test failures.
Actually I had that in my initial test patch [3], I don't know why I
missed that in final patch... I reran tests with the same approach of
[3] and found that there's still something missing, the
nmethod_entry_barrier() in c2 function prolog. The barrier may call to
runtime code (see generate_method_entry_barrier()). To reduce the risk
of missing such reinitialize_ptrue in newly added code in future, I
think it would be better to do the reinitialize in
pop_call_clobbered_registers().

P.S. the SIGILL message is also not clear, it should print detailed
message as indicated by MacroAssembler::stop() call. This is caused by
JDK-8255711 removing the message printing code. This will be fixed by JDK-8259539.

Tested with tier1-3 on SVE hardware. Also verified with the same
approach of patch [3] with jtreg tests hotspot_all_no_apps and
jdk:tier1-3 passed without incorrect ptrue value assertion failure.

[1] https://github.com/openjdk/jdk/pull/1621
[2] https://github.com/torvalds/linux/blob/master/Documentation/arm64/sve.rst
[3] http://cr.openjdk.java.net/~njian/8231441/0001-RFC-Block-one-caller-save-register-for-C2.patch

---------
### Progress

### Issue
 * [JDK-8258384](https://bugs.openjdk.java.net/browse/JDK-8258384): AArch64: SVE verify_ptrue fails on some tests

### Reviewers
 * [Andrew Dinn](https://openjdk.java.net/census#adinn) (@adinn - **Reviewer**)
 * [Nick Gasson](https://openjdk.java.net/census#ngasson) (@nick-arm - Committer)

### Download
`$ git fetch https://git.openjdk.java.net/jdk16 pull/50/head:pull/50`
`$ git checkout pull/50`

### DIFF
--- a/src/hotspot/os_cpu/linux_aarch64/os_linux_aarch64.cpp
+++ b/src/hotspot/os_cpu/linux_aarch64/os_linux_aarch64.cpp
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 1999, 2020, Oracle and/or its affiliates. All rights reserved.
+ * Copyright (c) 1999, 2021, Oracle and/or its affiliates. All rights reserved.
  * Copyright (c) 2014, Red Hat Inc. All rights reserved.
  * DO NOT ALTER OR REMOVE COPYRIGHT NOTICES OR THIS FILE HEADER.
  *
--- a/src/hotspot/os_cpu/linux_aarch64/os_linux_aarch64.cpp
+++ b/src/hotspot/os_cpu/linux_aarch64/os_linux_aarch64.cpp
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 1999, 2021, Oracle and/or its affiliates. All rights reserved.
+ * Copyright (c) 1999, 2020, Oracle and/or its affiliates. All rights reserved.
  * Copyright (c) 2014, Red Hat Inc. All rights reserved.
  * DO NOT ALTER OR REMOVE COPYRIGHT NOTICES OR THIS FILE HEADER.
  *
@@ -245,10 +245,7 @@ bool PosixSignals::pd_hotspot_signal_handler(int sig, siginfo_t* info,
           tty->print_cr("trap: %s: (SIGILL)", msg);
         }
 
-        va_list detail_args;
-        VMError::report_and_die(INTERNAL_ERROR, msg, detail_msg, detail_args, thread,
-                                pc, info, NULL, NULL, 0, 0);
-        va_end(detail_args);
+        return false; // Fatal error
       }
       else
 
--- a/src/hotspot/os_cpu/linux_aarch64/os_linux_aarch64.cpp
+++ b/src/hotspot/os_cpu/linux_aarch64/os_linux_aarch64.cpp
@@ -245,7 +245,10 @@ bool PosixSignals::pd_hotspot_signal_handler(int sig, siginfo_t* info,
           tty->print_cr("trap: %s: (SIGILL)", msg);
         }
 
-        return false; // Fatal error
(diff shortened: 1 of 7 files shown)


nsjian:
Can I get a review for jdk16 please?

I think the x86 pre-submit test failure is not related to this patch, which looks like the issue fixed by: https://github.com/openjdk/jdk16/pull/64

src/hotspot/os_cpu/linux_aarch64/os_linux_aarch64.cpp
```diff
@@ -245,7 +245,10 @@ bool PosixSignals::pd_hotspot_signal_handler(int sig, siginfo_t* info,
           tty->print_cr("trap: %s: (SIGILL)", msg);
         }
 
-        return false; // Fatal error
+        va_list detail_args;
+        VMError::report_and_die(INTERNAL_ERROR, msg, detail_msg, detail_args, thread,
+                                pc, info, NULL, NULL, 0, 0);
+        va_end(detail_args);
       }
```
adinn: I'm not sure it is ok to revert this code. The fix this is part of (for JDK-8255711) was provided explicitly to re-organize the flow of control for handling of fatal errors. Reverting this code appears to undermine the goal of that issue. I would like to get Thomas Stuefe's (@tstuefe) opinion on whether it is appropriate to abort the JVM here vs returning false before accepting this specific change.

adinn:
Hi Ningsheng,

The SVE changes look ok. I'm just unsure about restoring the exit in the signal handler.

nsjian:
> Hi Ningsheng,
> 
> The SVE changes look ok. I'm just unsure about restoring the exit in the signal handler.

Thank you @adinn  for the review! @tstuefe would you mind helping to review the lines @adinn mentioned? Thanks!

src/hotspot/os_cpu/linux_aarch64/os_linux_aarch64.cpp
tstuefe: Sorry for the delay. I'll take a look.

tstuefe: Thanks for finding that, and for pinging me. Please revert this part of the change, I opened a separate issue for it (also affects ppc64): https://bugs.openjdk.java.net/browse/JDK-8259539.

nsjian: Thank you Thomas. I have removed the change from my patch. Would you mind taking another look and approving this patch @adinn ?

adinn (approved):
This looks fine now.

nick-arm (approved):

nsjian:
Thank you @adinn @nick-arm  for the review!

/integrate

PR #53, opened 2020-12-18
8258007: Add instrumentation to NativeLibraryTest

bchristi-git:
This change adds some extra test output for NativeLibraryTest, primarily via an update to the ForceGC utility class.

It was observed that there was nothing preventing the Cleaner from cleaning the short-lived Object that ForceGC registers before await()/doit()/System.gc() is even called.

The new 'o' reference is kept alive until FoceGC.await() has been called.

We should find out a little more the next time NativeLibraryTest fails (or perhaps it won't fail anymore!)

---------
### Progress

### Issue
 * [JDK-8258007](https://bugs.openjdk.java.net/browse/JDK-8258007): Add instrumentation to NativeLibraryTest

### Reviewers
 * [Mandy Chung](https://openjdk.java.net/census#mchung) (@mlchung - **Reviewer**) ⚠️ Review applies to 54a9e6d19cbd2239c9b5475aa2b23c9bba014b26
 * [Naoto Sato](https://openjdk.java.net/census#naoto) (@naotoj - **Reviewer**) ⚠️ Review applies to 9b5e57ca6e4c0ab52e72190c16e62c770a1adf43

### Download
`$ git fetch https://git.openjdk.java.net/jdk16 pull/53/head:pull/53`
`$ git checkout pull/53`

### DIFF
--- a/test/lib/jdk/test/lib/util/ForceGC.java
+++ b/test/lib/jdk/test/lib/util/ForceGC.java
@@ -34,15 +34,18 @@ import java.util.function.BooleanSupplier;
 public class ForceGC {
     private final CountDownLatch cleanerInvoked = new CountDownLatch(1);
     private final Cleaner cleaner = Cleaner.create();
+    private Object o;
+
     public ForceGC() {
-        cleaner.register(new Object(), () -> cleanerInvoked.countDown());
+        this.o = new Object();
+        cleaner.register(o, () -> cleanerInvoked.countDown());
     }
 
-    private void doit() {
+    private void doit(int iter) {
         try {
             for (int i = 0; i < 10; i++) {
                 System.gc();
-                System.out.println("gc " + i);
+                System.out.format("doit %d: gc %d%n", iter, i);
                 if (cleanerInvoked.await(1L, TimeUnit.SECONDS)) {
                     return;
                 }
@@ -62,9 +65,10 @@ public class ForceGC {
      * @throws InterruptedException if the current thread is interrupted while waiting
      */
     public boolean await(BooleanSupplier s) {
+        o = null;
         for (int i = 0; i < 10; i++) {
             if (s.getAsBoolean()) return true;
-            doit();
+            doit(i);
         }
         return false;
     }
--- a/test/lib/jdk/test/lib/util/ForceGC.java
+++ b/test/lib/jdk/test/lib/util/ForceGC.java
@@ -45,7 +45,7 @@ public class ForceGC {
         try {
             for (int i = 0; i < 10; i++) {
                 System.gc();
-                System.out.format("doit %d: gc %d%n", iter, i);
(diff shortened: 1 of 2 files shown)


test/lib/jdk/test/lib/util/ForceGC.java
```diff
@@ -34,15 +34,18 @@
...
     }
 
-    private void doit() {
+    private void doit(int iter) {
         try {
             for (int i = 0; i < 10; i++) {
                 System.gc();
-                System.out.println("gc " + i);
+                System.out.format("doit %d: gc %d%n", iter, i);
                 if (cleanerInvoked.await(1L, TimeUnit.SECONDS)) {
```
mlchung, line 49: If the object that the test waits for garbage collected takes more GC cycles than the object registered in ForceGC's cleaner, this will return immediately.   If you are concerned, maybe if the count down latch already becomes zero, it should call System.sleep before the next System.gc().

mlchung (approved):
This looks okay.

test/lib/jdk/test/lib/util/ForceGC.java
```diff
@@ -34,15 +34,18 @@
...
+        cleaner.register(o, () -> cleanerInvoked.countDown());
     }
 
-    private void doit() {
+    private void doit(int iter) {
         try {
             for (int i = 0; i < 10; i++) {
                 System.gc();
-                System.out.println("gc " + i);
+                System.out.format("doit %d: gc %d%n", iter, i);
```
naotoj: Could be better to keep using println(), if performance is important.

test/lib/jdk/test/lib/util/ForceGC.java
bchristi-git: I am a little concerned.  But I think it makes more sense to sleep in await(), between calls of doit().  I'll push an update.

naotoj (approved):
Looks good to me.

test/lib/jdk/test/lib/util/ForceGC.java
bchristi-git: I agree we don't need format() here

test/lib/jdk/test/lib/util/ForceGC.java
```diff
@@ -62,9 +65,14 @@ private void doit() {
      * @throws InterruptedException if the current thread is interrupted while waiting
      */
     public boolean await(BooleanSupplier s) {
+        o = null; // Keep reference to Object until now, to ensure the Cleaner
+                  // doesn't count down the latch before await() is called.
         for (int i = 0; i < 10; i++) {
             if (s.getAsBoolean()) return true;
-            doit();
+            doit(i);
+            try { Thread.sleep(1000); } catch (InterruptedException e) {
```
mlchung, line 73: nit: can you break the catch clause in a separate line.  Otherwise looks good.

mlchung (approved):

PR #54, opened 2020-12-19
JDK-8258443: doclint should be service-loaded with system class loader

jonathan-gibbons:
Trivial fix to use the system class loader when loading doclint for javac

---------
### Progress

### Issue
 * [JDK-8258443](https://bugs.openjdk.java.net/browse/JDK-8258443): doclint should be service-loaded with system class loader

### Reviewers
 * [Alan Bateman](https://openjdk.java.net/census#alanb) (@AlanBateman - **Reviewer**)

### Download
`$ git fetch https://git.openjdk.java.net/jdk16 pull/54/head:pull/54`
`$ git checkout pull/54`

### DIFF
--- a/src/jdk.compiler/share/classes/com/sun/tools/doclint/DocLint.java
+++ b/src/jdk.compiler/share/classes/com/sun/tools/doclint/DocLint.java
@@ -50,7 +50,7 @@ public abstract class DocLint implements Plugin {
 
     public static synchronized DocLint newDocLint() {
         if (docLintProvider == null) {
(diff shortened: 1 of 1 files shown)


AlanBateman (approved):
Looks good, should allow JDK-8253996 to be closed too.

jonathan-gibbons:
@AlanBateman JDK-8253996 is slightly different; my intent is that javac/doclint should give a more friendly message when doclint is _really_ not available, such as in a minimal jlink-ed image.

PR #60, opened 2020-12-23
JDK-8253996: Javac error on jdk16 build 18: invalid flag: -Xdoclint:-missing

jonathan-gibbons:
This PR improves the diagnostic when javac is run in an image that does not provide a service provider for DocLint.

Instead of reporting messages about unsupported options, the following warning will be generated:

````
Warning: No service provider for doclint is available
````

In addition, as a minor cleanup, in the `jdk.javadoc` module, javadoc b bypasses the service loader to access `doclint` directly.

---------
### Progress

### Issue
 * [JDK-8253996](https://bugs.openjdk.java.net/browse/JDK-8253996): Javac error on jdk16 build 18: invalid flag: -Xdoclint:-missing

### Reviewers
 * [Hannes Wallnöfer](https://openjdk.java.net/census#hannesw) (@hns - **Reviewer**) ⚠️ Review applies to f9aacd5c8d5951aee31377d612de157ab356026c

### Download
`$ git fetch https://git.openjdk.java.net/jdk16 pull/60/head:pull/60`
`$ git checkout pull/60`

### CODE REVIEW

File: test/langtools/tools/javac/doclint/LimitedImage.java [hunk 1]
```diff
@@ -0,0 +1,75 @@
...
+        new JavacTask(tb, Mode.CMDLINE)
+                .options("--help")
+                .run().writeAll();
+
+        Path testSource = Path.of("Test.java");
+        tb.writeFile(testSource, "class Test {}");
+
+        List<String> actualOutput;
+        List<String> expectedOutput = List.of(
+                "- compiler.warn.doclint.not.available",
```
hns, line 58: Why is the warning key printed instead of the warning message? Limited image breaks property lookup?

test/langtools/tools/javac/doclint/LimitedImage.java
```diff
@@ -0,0 +1,75 @@
...
+                "1 warning"
+        );
+
+        //check proper diagnostics when zip/jar FS not present:
```
hns: Should be "doclint" or "javadoc" instead of "zip/jar FS"

hns (approved):
Looks good!

test/langtools/tools/javac/doclint/LimitedImage.java
jonathan-gibbons: It is common practice in jdk.compiler tests to use `-XDrawDiagnostics` which disables the use of resource files. The reason is to isolate the tests from the locale in which they may be executed, and from minor changes in the resources in the resource files, like spelling and grammar fixes, etc.

test/langtools/tools/javac/doclint/LimitedImage.java
jonathan-gibbons: I had to double check, but yes; thanks: good catch

PR #66, opened 2020-12-23
8253954: javac crash when compiling code with enhanced switch expressions with option -Xjcov

vicente-romero-oracle:
Please review this one liner patch that is fixing a crash in javac when compiling with option `-Xjcov`. The crash was due to an assertion failure at `com.sun.tools.javac.jvm.CRTable` as no position was found for compiler generated `yield` trees in some cases. The fix is to visit the `value` expression in the `yield` tree so that when the CRT table is being generated, the position for the `yield` tree is not `null`

TIA

---------
### Progress

### Issue
 * [JDK-8253954](https://bugs.openjdk.java.net/browse/JDK-8253954): javac crash when compiling code with enhanced switch expressions with option -Xjcov

### Reviewers
 * [Jonathan Gibbons](https://openjdk.java.net/census#jjg) (@jonathan-gibbons - **Reviewer**)

### Download
`$ git fetch https://git.openjdk.java.net/jdk16 pull/66/head:pull/66`
`$ git checkout pull/66`


jonathan-gibbons (approved):
Looks OK to me

PR #71, opened 2020-12-25
JDK-8249633: doclint reports missing javadoc for JavaFX property methods that have a property description

jonathan-gibbons:
There are two parts to the fix:

1. Similar to that of JDK-8251200:  predicate methods in `Utils` used to check if an element has certain tags in its doc comment may trigger the missing comment check even when there does not need to be a comment.

2. When synthesizing comments for the property-related methods, the code checks for the existence of explicit comments, in case they exist; that could also trigger the "missing comment" check.

The fix shares some changes with JDK-8251200; whichever does not go first will need a merge commit.

---------
### Progress

### Issue
 * [JDK-8249633](https://bugs.openjdk.java.net/browse/JDK-8249633): doclint reports missing javadoc for JavaFX property methods that have a property description

### Reviewers
 * [Hannes Wallnöfer](https://openjdk.java.net/census#hannesw) (@hns - **Reviewer**) ⚠️ Review applies to 39e1ec74346b889733cc1d4a0a62acacd45e6d26

### Download
`$ git fetch https://git.openjdk.java.net/jdk16 pull/71/head:pull/71`
`$ git checkout pull/71`


src/jdk.javadoc/share/classes/jdk/javadoc/internal/doclets/toolkit/util/Utils.java
```diff
@@ -2685,35 +2687,24 @@ public TreePath getTreePath(Element e) {
      */
     private final Map<Element, DocCommentInfo> dcTreeCache = new LinkedHashMap<>();
 
+    /**
+     * Checks whether an element has an associated doc copmment.
```
hns: typo: copmment

src/jdk.javadoc/share/classes/jdk/javadoc/internal/doclets/toolkit/util/Utils.java
```diff
@@ -2685,35 +2687,24 @@ public TreePath getTreePath(Element e) {
...
+     * @param element the element
+     * @return {@code true} if the eleme nt has a comment, and false otherwise
```
hns: typo: eleme nt

hns (approved):
Looks good, same two typos as in previous PR :-)

PR #72, opened 2020-12-29
8258972: unexpected compilation error with generic sealed interface

vicente-romero-oracle:
Hi,

Please review this fix to the implementation of section `5.1.6.1 Allowed Narrowing Reference Conversion` of the Sealed Classes spec [1](http://cr.openjdk.java.net/~gbierman/jep397/jep397-20201204/specs/sealed-classes-jls.html#jls-5.1.6.1). IMO the compiler was not implementing the spec to the letter in particular the inheritance relation between a sealed class and a subclass should be tested ignoring type arguments which is what this patch is doing. Comments?

Thanks,
Vicente

---------
### Progress

### Issue
 * [JDK-8258972](https://bugs.openjdk.java.net/browse/JDK-8258972): unexpected compilation error with generic sealed interface

### Reviewers
 * [Jan Lahoda](https://openjdk.java.net/census#jlahoda) (@lahodaj - **Reviewer**)

### Download
`$ git fetch https://git.openjdk.java.net/jdk16 pull/72/head:pull/72`
`$ git checkout pull/72`

### DIFF
--- a/src/jdk.compiler/share/classes/com/sun/tools/javac/code/Types.java
+++ b/src/jdk.compiler/share/classes/com/sun/tools/javac/code/Types.java
@@ -1667,7 +1667,7 @@ public class Types {
     }
     // where
         private boolean areDisjoint(ClassSymbol ts, ClassSymbol ss) {
-            if (isSubtype(ts.type, ss.type)) {
+            if (isSubtype(erasure(ts.type), erasure(ss.type))) {
                 return false;
             }
             // if both are classes or both are interfaces, shortcut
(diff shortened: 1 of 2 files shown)


src/jdk.compiler/share/classes/com/sun/tools/javac/code/Types.java
```diff
@@ -1667,7 +1667,7 @@ public boolean isCastable(Type t, Type s, Warner warn) {
     }
     // where
         private boolean areDisjoint(ClassSymbol ts, ClassSymbol ss) {
-            if (isSubtype(ts.type, ss.type)) {
+            if (isSubtype(erasure(ts.type), erasure(ss.type))) {
```
lahodaj, line 1670: There is another `isSubtype` below - could you please check if it needs the same erasure handling?

vicente-romero-oracle: I agree, done

lahodaj (approved):
Looks good to me.

PR #75, opened 2021-01-01
8258985: Parallel WeakProcessor may use too few threads

kimbarrett:
Please review this fix to the parallel WeakProcessor's computation of the
number of worker threads to use.  It was previously limited by the current
value of active_workers(), whatever that happens to be.  It should be
limited by total_workers(), just as with the parallel ReferenceProcessor.
(Both are subject to ReferencesPerThread.)

Testing
mach5 tier1
Some hand testing (Linux-x64) to verify the expected number of threads are
being used.

Note: That hand testing suggests some further tuning of ReferencesPerThread
might be in order. With the current default of 1000, I often saw in testing
that some threads were started late enough that no work was left for them.
I'll file a separate RFE for that.

/summary Use total workers rather than active.

---------
### Progress

### Issue
 * [JDK-8258985](https://bugs.openjdk.java.net/browse/JDK-8258985): Parallel WeakProcessor may use too few threads

### Reviewers
 * [Thomas Schatzl](https://openjdk.java.net/census#tschatzl) (@tschatzl - **Reviewer**) ⚠️ Review applies to 6859e5f68c4134ec8f67a0787c87f9efff28a57b
 * [Albert Mingkun Yang](https://openjdk.java.net/census#ayang) (@albertnetymk - Author) ⚠️ Review applies to 0ed088d8d73cc71f093a0e59f5e27d1ff4331f18
 * [Stefan Johansson](https://openjdk.java.net/census#sjohanss) (@kstefanj - **Reviewer**) ⚠️ Review applies to 0ed088d8d73cc71f093a0e59f5e27d1ff4331f18

### Download
`$ git fetch https://git.openjdk.java.net/jdk16 pull/75/head:pull/75`
`$ git checkout pull/75`


tschatzl (approved):
Lgtm although the comments in the documentation for both `static void weak_oops_do` method declarations mentions:

```
  // Parallel version.  Uses ergo_workers(), active workers, and
  // phase_time's max_threads to determine the number of threads to use.
```
which should be fixed.

One option is to put something like "Uses max workers and the total number of weak references to determine the number of threads to use" as description for `ergo_workers` and remove the details in the description of the `weak_oops_do` descriptions, but just removing the mention of `active_workers` there could be fine too.

kimbarrett:
Thanks @tschatzl . I've updated the descriptions of the weak_oops_do functions, and also added a description for ergo_workers.

albertnetymk:
What's the argument for using `total_workers()` here? BTW, `G1ConcurrentMark::weak_refs_work` is in the caller chain. According to its name, it happens outside a pause.

albertnetymk:
> weak_refs_work is called in the remark pause.

Indeed, I should have followed the call chain one more step. Then, `WeakProcessor::weak_oops_do` is always called in a pause, right? Maybe this is worth mentioning in the comment.

albertnetymk (approved):

tschatzl:
Still good. Thanks.

kstefanj (approved):

kimbarrett:
Thanks @tschatzl , @albertnetymk , @kstefanj for your reviews.

PR #83, opened 2021-01-05
8259027: NullPointerException in makeMappedSegment due to NULL Unmapper when length of segment is 0

mcimadamore:
When the size of the memory map is zero, FileChannelImpl returns a `null` Unmapper - this creates issues to the mapped memory segment implementation.

To fix, I've created an empty mapped segment class which is initialized to sensible defaults, and whose implenentation of force/load etc. do nothing.

We already had a test for this condition - but the test was missing the `@Test` annotation, so it was not run! I've now beefed up the test a bit to make sure that mapped segment operations do not throw.

---------
### Progress

### Issue
 * [JDK-8259027](https://bugs.openjdk.java.net/browse/JDK-8259027): NullPointerException in makeMappedSegment due to NULL Unmapper when length of segment is 0

### Reviewers
 * [Chris Hegarty](https://openjdk.java.net/census#chegar) (@ChrisHegarty - **Reviewer**)
 * [Uwe Schindler](https://openjdk.java.net/census#uschindler) (@uschindler - Author)

### Download
`$ git fetch https://git.openjdk.java.net/jdk16 pull/83/head:pull/83`
`$ git checkout pull/83`

### DIFF
--- a/src/jdk.incubator.foreign/share/classes/jdk/internal/foreign/MappedMemorySegmentImpl.java
+++ b/src/jdk.incubator.foreign/share/classes/jdk/internal/foreign/MappedMemorySegmentImpl.java
@@ -116,16 +116,16 @@ public class MappedMemorySegmentImpl extends NativeMemorySegmentImpl {
         if (bytesOffset < 0) throw new IllegalArgumentException("Requested bytes offset must be >= 0.");
         try (FileChannelImpl channelImpl = (FileChannelImpl)FileChannel.open(path, openOptions(mapMode))) {
             UnmapperProxy unmapperProxy = channelImpl.mapInternal(mapMode, bytesOffset, bytesSize);
+            int modes = defaultAccessModes(bytesSize);
+            if (mapMode == FileChannel.MapMode.READ_ONLY) {
+                modes &= ~WRITE;
+            }
             if (unmapperProxy != null) {
                 MemoryScope scope = MemoryScope.createConfined(null, unmapperProxy::unmap, null);
-                int modes = defaultAccessModes(bytesSize);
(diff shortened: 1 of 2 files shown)


src/jdk.incubator.foreign/share/classes/jdk/internal/foreign/MappedMemorySegmentImpl.java
```diff
@@ -116,13 +116,17 @@ public static MemorySegment makeMappedSegment(Path path, long bytesOffset, long
...
+            if (unmapperProxy != null) {
+                MemoryScope scope = MemoryScope.createConfined(null, unmapperProxy::unmap, null);
+                int modes = defaultAccessModes(bytesSize);
+                if (mapMode == FileChannel.MapMode.READ_ONLY) {
+                    modes &= ~WRITE;
+                }
+                return new MappedMemorySegmentImpl(unmapperProxy.address(), unmapperProxy, bytesSize,
+                        modes, scope);
+            } else {
+                return new EmptyMappedMemorySegmentImpl();
```
ChrisHegarty: While a little pedantic, should the `mapMode` be considered so that an empty mapped segment will feature the expected access modes when queried through the `accessModes` API ?

mcimadamore: Ugh - yes, good point

ChrisHegarty (approved):
Looks good to me.

uschindler (approved):
Looks fine!
In fact it's more complicated than the workaround, but this allows to call load() or force().

PR #84, opened 2021-01-05
8259032: MappedMemorySegmentImpl#makeMappedSegment() ignores Unmapper#pagePosition

mcimadamore:
Hi,
this patch fixes an oversight where the address() method in FileChannelImpl.Unmapper does not take into account the value of `pagePosition` - this leads to a situation in which, effectively, the specified offset when mapping the segment is ignored by the runtime (because base address will always be aligned to some known quantity - which is OS/platform dependent).

To test this I had to break open into FileChannelImpl and ready the granularity.

---------
### Progress

### Issue
 * [JDK-8259032](https://bugs.openjdk.java.net/browse/JDK-8259032): MappedMemorySegmentImpl#makeMappedSegment() ignores Unmapper#pagePosition

### Reviewers
 * [Alan Bateman](https://openjdk.java.net/census#alanb) (@AlanBateman - **Reviewer**) ⚠️ Review applies to 7a78d72474f221578c30374c3d07fa91e992c5c2

### Contributors
 * Uwe Schindler `<uschindler@openjdk.org>`

### Download
`$ git fetch https://git.openjdk.java.net/jdk16 pull/84/head:pull/84`
`$ git checkout pull/84`


test/jdk/java/foreign/TestByteBuffer.java
```diff
@@ -25,7 +25,7 @@
  * @test
  * @modules java.base/sun.nio.ch
  *          jdk.incubator.foreign/jdk.internal.foreign
- * @run testng/othervm -Dforeign.restricted=permit TestByteBuffer
+ * @run testng/othervm --illegal-access=permit -Dforeign.restricted=permit TestByteBuffer
```
AlanBateman: Can you change java.base/sun.nio.ch to java.base/sun.nio.ch:+open instead? That would avoid the --illegal-access=permit.

uschindler: I am not sure if a test like this is really needed.
The alignment is pageSize on Linux and some arbitrary value (65536) on Windows. If you have some test file that writes like a few bytes (1, 2, 3, 4,...) To a file and then maps it with offsets other than zero, you just have to get the first byte and compare it to offset.

uschindler: In fact the new test only checks if everything is aligned like we expect, but it does not test that our mapping returns a memory segment with expected contents.

uschindler: So I tend to make a simple test without reflection that writes a defined sequence of bytes to a file and then maps it with various offsets, checking that the first byte in the mapped segment is from that sequence.

mcimadamore: Good suggestion!

uschindler:
Hi,
The new test looks fine to me.
Uwe

AlanBateman (approved):
Good suggestion from Uwe.

PR #89, opened 2021-01-06
8259227: C2 crashes with SIGFPE due to a division that floats above its zero check

chhagedorn:
This bug is very similar to [JDK-8257822](https://bugs.openjdk.java.net/browse/JDK-8257822). In this testcase, a `Div` node has no longer its zero check as direct control input and is later moved before the zero check by `IfNode::dominated_by()` which updates all data nodes to a dominating `If` (in JDK-8257822 it was done by `PhaseIdealLoop::dominated_by()`) .

I suggest to use the same fix for `IfNode::dominated_by()` as for `PhaseIdealLoop::dominated_by()` in JDK-8257822 to only move data nodes to the dominating `If` if it is not a `Div` or `Mod` node that could have a zero divisor (i.e. a zero check).

Thanks,
Christian

---------
### Progress

### Issue
 * [JDK-8259227](https://bugs.openjdk.java.net/browse/JDK-8259227): C2 crashes with SIGFPE due to a division that floats above its zero check

### Reviewers
 * [Vladimir Kozlov](https://openjdk.java.net/census#kvn) (@vnkozlov - **Reviewer**) ⚠️ Review applies to 4273fcfba59d1826a9bf51a9a6de0e49f3393686

### Download
`$ git fetch https://git.openjdk.java.net/jdk16 pull/89/head:pull/89`
`$ git checkout pull/89`

### DIFF
--- a/test/hotspot/jtreg/compiler/loopopts/TestDivZeroDominatedBy.java
+++ b/test/hotspot/jtreg/compiler/loopopts/TestDivZeroDominatedBy.java
@@ -24,11 +24,14 @@
 
 /*
  * @test
+ * @key stress randomness
  * @bug 8259227
  * @summary Verify that zero check is executed before division/modulo operation.
  * @requires vm.compiler2.enabled
  * @run main/othervm -Xcomp -XX:-TieredCompilation -XX:CompileOnly=compiler/loopopts/TestDivZeroDominatedBy::test
  *                   -XX:+UnlockDiagnosticVMOptions -XX:+StressGCM -XX:StressSeed=917280111 compiler.loopopts.TestDivZeroDominatedBy
+ * @run main/othervm -Xcomp -XX:-TieredCompilation -XX:CompileOnly=compiler/loopopts/TestDivZeroDominatedBy::test
+ *                   -XX:+UnlockDiagnosticVMOptions -XX:+StressGCM compiler.loopopts.TestDivZeroDominatedBy
  */
 
 package compiler.loopopts;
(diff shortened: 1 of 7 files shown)


vnkozlov (approved):
Looks good.

chhagedorn:
Thanks for your review Vladimir!

test/hotspot/jtreg/compiler/loopopts/TestDivZeroDominatedBy.java
```diff
@@ -0,0 +1,75 @@
...
+ *
+ */
+
+/*
+ * @test
+ * @bug 8259227
+ * @summary Verify that zero check is executed before division/modulo operation.
+ * @requires vm.compiler2.enabled
+ * @run main/othervm -Xcomp -XX:-TieredCompilation -XX:CompileOnly=compiler/loopopts/TestDivZeroDominatedBy::test
+ *                   -XX:+UnlockDiagnosticVMOptions -XX:+StressGCM -XX:StressSeed=917280111 compiler.loopopts.TestDivZeroDominatedBy
```
TobiHartmann, line 32: Just wondering if the `StressSeed` should really be hardcoded? If things change in the VM, this seed might not longer reproduce the issue. Maybe add another `@run `with no fixed seed?

TobiHartmann (approved):
Otherwise looks good to me.

test/hotspot/jtreg/compiler/loopopts/TestDivZeroDominatedBy.java
chhagedorn: Good idea. I also updated the test from JDK-8257822 with such an additional run.

TobiHartmann (approved):
Thanks for updating, looks good!

chhagedorn:
Thanks Tobias for your review!

PR #90, opened 2021-01-06
8259028: ClassCastException when using custom filesystem with wrapper FileChannel impl

mcimadamore:
This patch tweaks `MemorySegment::mapFile` so that it will throw `IllegalArgumentException` whenever the path to be mapped is associated with a custom file system provider.

The check in the implementation is heavily borrowed by what `UnixDomainSocketAddress::of(Path)` does (thanks Alan for the tip!). Not only we have to check if file system is the default one, but also if the default FS belongs to java.base (since that can be overridden).

The test simply check that paths coming from the (internal) JRT file system are not supported by the factory.

---------
### Progress

### Issue
 * [JDK-8259028](https://bugs.openjdk.java.net/browse/JDK-8259028): ClassCastException when using custom filesystem with wrapper FileChannel impl

### Reviewers
 * [Chris Hegarty](https://openjdk.java.net/census#chegar) (@ChrisHegarty - **Reviewer**)
 * [Alan Bateman](https://openjdk.java.net/census#alanb) (@AlanBateman - **Reviewer**)
 * [Uwe Schindler](https://openjdk.java.net/census#uschindler) (@uschindler - Author)

### Download
`$ git fetch https://git.openjdk.java.net/jdk16 pull/90/head:pull/90`
`$ git checkout pull/90`

### DIFF
--- a/src/jdk.incubator.foreign/share/classes/jdk/incubator/foreign/MemorySegment.java
+++ b/src/jdk.incubator.foreign/share/classes/jdk/incubator/foreign/MemorySegment.java
@@ -876,10 +876,9 @@ allocateNative(bytesSize, 1);
      * @param mapMode a file mapping mode, see {@link FileChannel#map(FileChannel.MapMode, long, long)}; the chosen mapping mode
      *                might affect the behavior of the returned memory mapped segment (see {@link MappedMemorySegments#force(MemorySegment)}).
      * @return a new confined mapped memory segment.
-     * @throws IllegalArgumentException if {@code bytesOffset < 0}.
-     * @throws IllegalArgumentException if {@code bytesSize < 0}.
-     * @throws UnsupportedOperationException if an unsupported map mode is specified, or if the {@code path} is associated
-     * with a provider that does not support creating file channels.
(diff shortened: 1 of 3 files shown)


mcimadamore:
CSR: https://bugs.openjdk.java.net/browse/JDK-8259321

ChrisHegarty (approved):

AlanBateman (approved):

uschindler (approved):
Looks fine to me.

Another option would have been to just check the class with instanceof. If the FileChannel returned by the path's filesystem has wrong type, it won't work. If we have some custom filesystem that just returns the original FileChannelImpl but does some additional checks before or after (e.g to hide some files from user), this would still work.

But I think it's better to be clear here and deny anything which is not default operating system filesystem, until we have a better solution.

In Lucene I added some hack that removes our test-only filesystem wrappers from the path. I will later go the route to mark those path instances with some interface Unwrappable that provides a method unwrap().

mcimadamore:
> Another option would have been to just check the class with instanceof. If the FileChannel returned by the path's filesystem has wrong type, it won't work. If we have some custom filesystem that just returns the original FileChannelImpl but does some additional checks before or after (e.g to hide some files from user), this would still work.

We considered this - but we opted for better clarity and alignment with existing APIs.

PR #99, opened 2021-01-09
8259237: Demo selection changes with left/right arrow key. No need to press space for selection.

pankaj-bansal:
Please review a fix for jdk16

Issue: In SwingSet2, if the user navigates through the demos, the demo gets selected/starts on pressing left/right key only. There is no need to press the "space" key. Earlier, on pressing the left/right key, only demo icon used to get focused and user needed to press the "space" to actually select a demo.

Cause: The SwingSet2 has JToggleButtons added to a ButtonGroup. Each JToggleButton has an AbstractAction set on it, which loads the demo. Earlier, when the user pressed Left/Right button, only the selected button used to change. The Action set on JToggleButton used to perform only  on pressing the "Space" button. Now, the Action is performed on navigating the JToggleButtons using Left/Right keys without the need to press the "space" key.

This issue is a side effect of fix done for https://bugs.openjdk.java.net/browse/JDK-8249548. The issues fixed in JDK-8249548 were not present in JRadioButton as there was code available to handle it in AquaButtonRadioUI and BasicRadioButtonUI. The fix done for JDK-8249548 moved duplicate code from AquaButtonRadioUI and BasicRadioButtonUI and moved it to BasicButtonUI, so this code is available for JToggleButton and JRadioButton for all L&Fs. This was wrong as there is a difference in behaviour for JRadioButtons added to ButtonGroup for AquaL&F and other L&Fs. In AquaL&F, the AbstractAction set on JRadioButton is not performed on button selection and user has to press "space". In other L&Fs, the AbstractAction is performed on selection of Button itself without the need to press "space". 

Fix: The current change fixes the issue in present bug and keeps the fixes done in JDK-8249548. Following is the behaviour of JToggleButton and JRadioButton for different L&Fs before JDK-8249548 fix, after JDK-8249548 fix and after current fix.

Before fix for JDK-8249548
JToggleButton: 
For all L&Fs, user can not navigate/select the buttons added to ButtonGroup properly as mentioned in the JDK-8249548.
JRadioButton: 
For Synth L&F (Nimbus L&F), user can not navigate/select the buttons added to ButtonGroup. 
For AquaL&F, user can select the buttons added to ButtonGroup by pressing left/right key and needs to press the "space" to perform the set Action.
For Other L&Fs, user can select/navigate the buttons and the set Action is also performed without pressing "space"

After fix for JDK-8249548
JToggleButton:
For all L&Fs, user can navigate/select the buttons added to ButtonGroups and the AbstractAction set is performed without the need to press "space".
JRadioButton:
For all L&Fs, user can navigate/select the buttons added to ButtonGroups and the AbstractAction set is performed without the need to press "space".

After current fix:
JToggleButton:
For all L&Fs, user can navigate/select the buttons added to ButtonGroups.  User needs to press "space" to perform the set AbstractAction.
JRadioButton:
For AquaL&F, the behaviour before JDK-8249548 is restored, so user needs to press the "space" to perform the set Action.
For all other L&Fs (including Synth L&F), the behaviour is same. User can navigate/select the buttons added to ButtonGroups and set Action is performed without pressing "space"

I have run mach5 jobs with full jtreg/jck and all looks good. Links in JBS. The test TestButtonGroupFocusTraversal.java is modified such that it fails without current fix and passes after the fix. The fix also should be verified by running SwingSet2.

---------
### Progress

### Issue
 * [JDK-8259237](https://bugs.openjdk.java.net/browse/JDK-8259237): Demo selection changes with left/right arrow key. No need to press space for selection.

### Reviewers
 * [Prasanta Sadhukhan](https://openjdk.java.net/census#psadhukhan) (@prsadhuk - **Reviewer**) ⚠️ Review applies to 6c745600ad9b035276c5d76a4d3951c42c426229
 * [Alexander Zuev](https://openjdk.java.net/census#kizune) (@azuev-java - **Reviewer**) ⚠️ Review applies to de3db58ac1b9f686c02adb94f0378cf5618a78c8
 * [Sergey Bylokhov](https://openjdk.java.net/census#serb) (@mrserb - **Reviewer**)

### Download
`$ git fetch https://git.openjdk.java.net/jdk16 pull/99/head:pull/99`
`$ git checkout pull/99`


src/java.desktop/macosx/classes/com/apple/laf/AquaButtonUI.java
```diff
@@ -217,10 +221,15 @@ protected void uninstallKeyboardActions(final AbstractButton b) {
     }
 
     protected void uninstallListeners(final AbstractButton b) {
-        super.uninstallListeners(b);
         final AquaButtonListener listener = (AquaButtonListener)b.getClientProperty(this);
         b.putClientProperty(this, null);
         if (listener != null) {
+            b.removeMouseListener(listener);
+            b.removeMouseListener(listener);
```
prsadhuk: Why are we removing same listener twice?

prsadhuk (commented):
I guess you did not run jtreg job with latest problemlist as it did not run
javax/swing/JRadioButton/ButtonGroupFocus/ButtonGroupFocusTest.java
javax/swing/JRadioButton/8075609/bug8075609.java
javax/swing/JRadioButton/8033699/bug8033699.java
Please see this test pass with your change as some of them used to fail  after JDK-8226892 which was reworked by fix of  JDK-8249548

src/java.desktop/macosx/classes/com/apple/laf/AquaButtonUI.java
pankaj-bansal: I just rolled back the changes I did in this file in JDK-8249548. Looks like this was has been there for some time. See the file in commit made in this file before JDK-8249548 https://github.com/openjdk/jdk/blob/c18080fef714949221c8ddabc4e23d409575c3d3/src/java.desktop/macosx/classes/com/apple/laf/AquaButtonUI.java. 
But yes, it should not be done and I will remove it and update the PR.

azuev-java:
Quick question: why we special casing JRadioButton case? Is there any reason we should call the action on them?

pankaj-bansal:
> Quick question: why we special casing JRadioButton case? Is there any reason we should call the action on them?

This was the behaviour which was there before JDK-8249548 for JRadioButton and I am just reverting to it. As explained in the PR description, the behaviour before JDK-8249548 was as below

Before fix for JDK-8249548
JToggleButton:
    For all L&Fs, user can not navigate/select the buttons added to ButtonGroup properly as mentioned in the JDK-8249548.
JRadioButton:
    For Synth L&F (Nimbus L&F), user can not navigate/select the buttons added to ButtonGroup.
    For AquaL&F, user can select the buttons added to ButtonGroup by pressing left/right key and needs to press the "space" to perform the set Action.
    For Other L&Fs, user can select/navigate the buttons and the set Action is also performed without pressing "space"

pankaj-bansal:
I have removed the duplicate code for AquaL&F. I think during the fix made for JDK-8226892, it was a mistake to not change the AquaL&F code. So, the behaviour of JRadioButton should be similar across L&Fs and the duplicate code in AquaL&F is not needed. A mach5 job with all jtreg/jck tests is being run and I will will update when it finishes.

azuev-java (approved):
Looks ok now.

pankaj-bansal:
> I guess you did not run jtreg job with latest problemlist as it did not run
> javax/swing/JRadioButton/ButtonGroupFocus/ButtonGroupFocusTest.java
> javax/swing/JRadioButton/8075609/bug8075609.java
> javax/swing/JRadioButton/8033699/bug8033699.java
> Please see this test pass with your change as some of them used to fail after JDK-8226892 which was reworked by fix of JDK-8249548

I have updated the PR with some commits and ran the full jtreg/jck tests including the tests you have mentioned. The fix is not breaking anything.

test/jdk/javax/swing/ButtonGroup/TestButtonGroupFocusTraversal.java
```diff
@@ -30,6 +30,7 @@
  * @run main TestButtonGroupFocusTraversal
  */
```
prsadhuk, line 32: We need to append this bugid as product code is changed for which this testcase is modified

pankaj-bansal: Done, please have a look.

prsadhuk (approved):

mrserb:
The spec of the update method seems outdated:
         * Find the new toggle button that focus needs to be
         * moved to in the group, select the button

BTW I am not sure why the Aqua is so specific and requires an additional press while the same radio buttons do not require that. Is it possible that some fix was implemented in the shared code so now all L&F skip the "space"?

pankaj-bansal:
> The spec of the update method seems outdated:
> * Find the new toggle button that focus needs to be
> * moved to in the group, select the button
> 
> BTW I am not sure why the Aqua is so specific and requires an additional press while the same radio buttons do not require that. Is it possible that some fix was implemented in the shared code so now all L&F skip the "space"?

The AquaL&F also does not need the additional "space" press for JRadioButton. During the fix of JDK-8226892, code was added for other L&F, so that the additional "space" press is not needed. But same code was not added for AquaL&F. So, the additional key press was needed. Now, the additional press should not be needed for AquaL&F also for JRadioButton. So now, for all L&Fs, the additional key press is needed for JToggleButton and is not needed for JRadioButton.

I did not add radio button to spec as RadioButton are also ToggleButton only. But yes, I will highlight that. So I will update the spec as below. Hope that is ok.
         * Find the new toggle/radio button that focus needs to be
         * moved to in the group, select the button

mrserb:
> I did not add radio button to spec as RadioButton are also ToggleButton only. But yes, I will highlight that. So I will update the spec as below. Hope that is ok.
> * Find the new toggle/radio button that focus needs to be
> * moved to in the group, select the button

Also please add a comment of why we need to use the setPressed/setArmed it seems unrelated to the selection of the button, and is not described anywhere.

pankaj-bansal:
> > I did not add radio button to spec as RadioButton are also ToggleButton only. But yes, I will highlight that. So I will update the spec as below. Hope that is ok.
> > 
> > * Find the new toggle/radio button that focus needs to be
> > * moved to in the group, select the button
> 
> Also please add a comment of why we need to use the setPressed/setArmed it seems unrelated to the selection of the button, and is not described anywhere.

Done, please have a look.

test/jdk/javax/swing/ButtonGroup/TestButtonGroupFocusTraversal.java
```diff
@@ -76,6 +79,20 @@ public void run() {
                 radioButton1 = new JRadioButton("1");
                 radioButton2 = new JRadioButton("2");
```
mrserb, line 83: Just to be safe please check the JCheckBox, it also extends the JToggleButton and we should not break it.

mrserb:
Just small summary info for the future: we have a JToggleButton class and two subclasses: JCheckBox and JRadioButton.
1. It was decided(but not strictly specified) to generate an action when the selection is changed by the arrow keys for JRadioButton. It is just common sense.
2. It was decided to not generate an action when the selection is changed by arrow keys for JCheckBox. Since the user should have the ability to move over different checkboxes without toggling them.
3. The parent class JToggleButton is a root cause of the current bug under discussion. We could implement it in any way, we can generate an action when the user presses the arrow key, but the opposite just a safer solution now.

mrserb (approved):

PR #102, opened 2021-01-11
8259560: Zero m68k: "static assertion failed: align" after JDK-8252049

shipilev:
Debian folks report that Zero m68k (and seems only that arch) fails the static assert added by JDK-8252049. It affects at least JDK 16, but would affect more releases once JDK-8252049 is backported.

```
/<<PKGBUILDDIR>>/src/hotspot/share/oops/methodData.hpp: In constructor 'MethodData::CompilerCounters::CompilerCounters()':
/<<PKGBUILDDIR>>/src/hotspot/share/oops/methodData.hpp:1987:55: error: static assertion failed: align
 1987 | static_assert(sizeof(_trap_hist) % HeapWordSize == 0, "align");
      |
```

I tried to massage the code so that `oops/methodData.hpp` is not touched when Zero is built (since Zero does not use compilers and related infra), but there are enough transitive dependencies to `oops/methodData.hpp` to make it quite messy. So instead I would just special-case it in place.

---------
### Progress

### Issue
 * [JDK-8259560](https://bugs.openjdk.java.net/browse/JDK-8259560): Zero m68k: "static assertion failed: align" after JDK-8252049

### Reviewers
 * [David Holmes](https://openjdk.java.net/census#dholmes) (@dholmes-ora - **Reviewer**)

### Download
`$ git fetch https://git.openjdk.java.net/jdk16 pull/102/head:pull/102`
`$ git checkout pull/102`

### DIFF
--- a/src/hotspot/share/ci/ciMethodData.cpp
+++ b/src/hotspot/share/ci/ciMethodData.cpp
@@ -202,7 +202,12 @@ void ciMethodData::load_data() {
   // _extra_data_size = extra_data_limit - extra_data_base
   // total_size = _data_size + _extra_data_size
   // args_data_limit = data_base + total_size - parameter_data_size
+
+#ifndef ZERO
+  // Some Zero platforms do not have expected alignment, and do not use
+  // this code. static_assert would still fire and fail for them.
   static_assert(sizeof(_orig) % HeapWordSize == 0, "align");
+#endif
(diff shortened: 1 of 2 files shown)


dholmes-ora (approved):
I prefer NOT_ZERO macro usage over ifdefs, but don't insist.

Thanks,
David

glaubitz:
I can confirm that the current PR fixes the problem for me.

PR #104, opened 2021-01-11
8259275: JRuby crashes while resolving invokedynamic instruction

calvinccheung:
Please review this proposed change which fixes 2 problems during CDS dump time:

- in `ClassListParser::resolve_indy()`, exception could be thrown from `bootstrap_specifier.resolve_bsm()` but was not handled;
- in `ConstantPool::remove_unshareable_info()`, tag was not setup correctly on error conditions.

Passed tiers 1 - 4 testing.

---------
### Progress

### Issue
 * [JDK-8259275](https://bugs.openjdk.java.net/browse/JDK-8259275): JRuby crashes while resolving invokedynamic instruction

### Reviewers
 * [Ioi Lam](https://openjdk.java.net/census#iklam) (@iklam - **Reviewer**)
 * [Yumin Qi](https://openjdk.java.net/census#minqi) (@yminqi - **Reviewer**) ⚠️ Review applies to 6786de937f2ac820ca04dfe895c9139532504e9d
 * [Lois Foltan](https://openjdk.java.net/census#lfoltan) (@lfoltan - **Reviewer**) ⚠️ Review applies to e139e9646dd10bb268334c8a5c9f99725aa5f63c

### Download
`$ git fetch https://git.openjdk.java.net/jdk16 pull/104/head:pull/104`
`$ git checkout pull/104`

### DIFF
--- a/src/hotspot/share/classfile/classListParser.cpp
+++ b/src/hotspot/share/classfile/classListParser.cpp
@@ -474,10 +474,8 @@ void ClassListParser::resolve_indy(Symbol* class_name_symbol, TRAPS) {
       tty->print_cr(" exception pending %s ",
                     PENDING_EXCEPTION->klass()->external_name());
     }
-    Symbol* ex_name = PENDING_EXCEPTION->klass()->name();
     oop exception = THREAD->pending_exception();
-    if (/*ex_name == vmSymbols::java_lang_UnsupportedClassVersionError() ||*/
-        exception->is_a(SystemDictionary::Error_klass())) {
+    if (exception->is_a(SystemDictionary::Error_klass())) {
       CLEAR_PENDING_EXCEPTION;
     } else {
       exit(1);
--- a/src/hotspot/share/classfile/classListParser.cpp
+++ b/src/hotspot/share/classfile/classListParser.cpp
@@ -476,6 +476,8 @@ void ClassListParser::resolve_indy(Symbol* class_name_symbol, TRAPS) {
     }
     oop exception = THREAD->pending_exception();
     if (exception->is_a(SystemDictionary::Error_klass())) {
+      // Clear the exception associated with errors like UnsupportedClassVersionError
+      // or NoSuchMethodError so that CDS dumping can continue.
       CLEAR_PENDING_EXCEPTION;
     } else {
       exit(1);
--- a/src/hotspot/share/classfile/classListParser.cpp
+++ b/src/hotspot/share/classfile/classListParser.cpp
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2015, 2020, Oracle and/or its affiliates. All rights reserved.
+ * Copyright (c) 2015, 2021, Oracle and/or its affiliates. All rights reserved.
  * DO NOT ALTER OR REMOVE COPYRIGHT NOTICES OR THIS FILE HEADER.
  *
  * This code is free software; you can redistribute it and/or modify it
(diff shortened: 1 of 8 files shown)


src/hotspot/share/classfile/classListParser.cpp
```diff
@@ -460,19 +460,40 @@ bool ClassListParser::is_matching_cp_entry(constantPoolHandle &pool, int cp_inde
...
+      char* ex_msg = java_lang_String::as_utf8_string(message);
+      tty->print_cr(" exception pending '%s %s'",
+                    PENDING_EXCEPTION->klass()->external_name(), ex_msg);
+    } else {
+      tty->print_cr(" exception pending %s ",
+                    PENDING_EXCEPTION->klass()->external_name());
+    }
+    oop exception = THREAD->pending_exception();
+    if (exception->is_a(SystemDictionary::Error_klass())) {
+      CLEAR_PENDING_EXCEPTION;
```
lfoltan: Just for my clarification, why do you clear the pending exception in the case of Error_klass()?  Please consider adding a comment explaining this.

calvinccheung: Hi Lois,

Thanks for your review. I've added some comment in `ClassListParser::resolve_indy` for the `Error_klass` case.

iklam (approved):
Looks good to me.

yminqi (approved):
Looks good!

calvinccheung:
@iklam, @yminqi Thanks for your review.

src/hotspot/share/classfile/classListParser.cpp
```diff
@@ -460,19 +460,42 @@ bool ClassListParser::is_matching_cp_entry(constantPoolHandle &pool, int cp_inde
   }
   return true;
 }
-
 void ClassListParser::resolve_indy(Symbol* class_name_symbol, TRAPS) {
+  ClassListParser::resolve_indy_impl(class_name_symbol, THREAD);
+  if (HAS_PENDING_EXCEPTION) {
+    ResourceMark rm(THREAD);
+    tty->print("resolve_indy for class %s has", class_name_symbol->as_C_string());
```
dholmes-ora: Shouldn't this be using appropriate unified logging?

src/hotspot/share/classfile/classListParser.cpp
dholmes-ora: I would have expected only LinkageErrors to be ignored as they are the only "expected" errors aren't they?

calvinccheung: LinkageError should work since BootstrapMethodError, NoSuchMethodError, etc. are subclasses.
I've changed the check to:
`if (exception->is_a(SystemDictionary::LinkageError_klass())) {`

src/hotspot/share/classfile/classListParser.cpp
calvinccheung: Ok. I've changed it to `LogTarget(Debug, cds, lambda)`.

src/hotspot/share/classfile/classListParser.cpp
```diff
@@ -460,19 +461,46 @@ bool ClassListParser::is_matching_cp_entry(constantPoolHandle &pool, int cp_inde
...
+                  PENDING_EXCEPTION->klass()->external_name());
+      }
+    }
+    oop exception = THREAD->pending_exception();
+    if (exception->is_a(SystemDictionary::LinkageError_klass())) {
+      // Clear the exception associated with errors like UnsupportedClassVersionError
+      // or NoSuchMethodError so that CDS dumping can continue.
+      CLEAR_PENDING_EXCEPTION;
+    } else {
+      exit(1);
```
dholmes-ora: We should never just use `exit()` - at a minimum `vm_exit()`. But we should not silently exit with no indication of what went wrong. I'm not sure of the call chain here but can't we just return and let higher layers do the same until we eventually hit a cleaner exit path that reports the exception?

Edit: I see now that this code was moved from another location not introduced with this fix, but this still seems wrong to me.

dholmes-ora (commented):
I missed the fact some of this code was just being moved around when I took the first look, so my comments could be seen as something suitable for a follow up RFE.

This is not a full review.
Thanks,
David

src/hotspot/share/classfile/classListParser.cpp
calvinccheung: For now, I think the logging in ClassListParser::resolve_indy should just by tty->print instead of using UL. This way, if the exit(1) is hit, there will be some information about the reason. Also, not too many users would enable -Xlog:cds+lambda=debug.

lfoltan (approved):
Looks good, thank you for adding the comment!
Lois

src/hotspot/share/classfile/classListParser.cpp
calvinccheung: After offline discussion with Ioi, we've decided to use `log_warning(cds)` to output the exception message since log_warning does not require setting a log level. Also, the `exit(1)` is not needed, we can continue with the CDS dumping after printing the message and clearing the exception.

iklam (approved):

PR #105, opened 2021-01-11
8258272: LoadVectorMaskedNode can't be replaced by zero con

neliasso:
A test fails for this code:

byte [] src = new byte[0];
byte [] dest = Arrays::copy_of(src);

This is transformed to something like:
byte [] src = new byte[0];
byte [] dest = new byte[0];
System.arraycopy(src, 0, dest, 0, src.lenght());

This causes a problem in LoadNode::Value for the new LoadVectorMasked nodes. LoadNode::Value sees that the load will always load a zero - and will try to replace itself with a zero constant. That doesn't work for the LoadVectorMaskedNodes currently.

This patch adds a check to the zero-optimization to pass on vector types.

This patch is currently missing a new test. Since we are nearing the end of bugfixing in JDK16 I decided to publish the fix first so that any discussion of the fix can be had now.

---------
### Progress

### Issue
 * [JDK-8258272](https://bugs.openjdk.java.net/browse/JDK-8258272): LoadVectorMaskedNode can't be replaced by zero con

### Reviewers

### Download
`$ git fetch https://git.openjdk.java.net/jdk16 pull/105/head:pull/105`
`$ git checkout pull/105`

### DIFF
--- a/src/hotspot/share/opto/memnode.cpp
+++ b/src/hotspot/share/opto/memnode.cpp
@@ -2045,7 +2045,7 @@ const Type* LoadNode::Value(PhaseGVN* phase) const {
     }
   }
 
-  if (is_instance && !_type->is_vect()) {
+  if (is_instance && (_type->isa_vect() == NULL)) {
     // If we have an instance type and our memory input is the
     // programs's initial memory state, there is no matching store,
     // so just return a zero of the appropriate type -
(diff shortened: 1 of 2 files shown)


chhagedorn (approved):
That looks good to me!

iwanowww (approved):
Looks good.

chhagedorn (approved):
Looks good!

iwanowww (approved):
Even better!

chhagedorn (approved):

PR #107, opened 2021-01-12
8231461: static/instance overload leads to 'unexpected static method found in unbound lookup' when resolving method reference

vicente-romero-oracle:
This patch fixes the following issue, for code like:
```
import java.util.function.*;
 
public class Test {
    public String foo(Object o) { return "foo"; }
    public static String foo(String o) { return "bar"; }

    public void test() {
        Function<String, String> f = Test::foo;
    }
}
```
javac is issuing an invalid method reference error even though the static method is applicable. This is due to a regression on the implementation of the following assertion on section 15.13.1 of the JLS 14:

```
If the first search produces a most specific method that is static , and the set
of applicable methods produced by the second search contains no non- static
methods, then the compile-time declaration is the most specified method of
the first search.
```

Apart from fixing the mentioned regression, this patch also adds a debug feature for the compiler to print out the findings during the search for an applicable method reference. The information is printed out if a debug flag is passed to the compiler.

TIA

---------
### Progress

### Issue
 * [JDK-8231461](https://bugs.openjdk.java.net/browse/JDK-8231461): static/instance overload leads to 'unexpected static method found in unbound lookup' when resolving method reference

### Reviewers
 * [Maurizio Cimadamore](https://openjdk.java.net/census#mcimadamore) (@mcimadamore - **Reviewer**)

### Download
`$ git fetch https://git.openjdk.java.net/jdk16 pull/107/head:pull/107`
`$ git checkout pull/107`


src/jdk.compiler/share/classes/com/sun/tools/javac/comp/Resolve.java
```diff
@@ -3280,12 +3324,12 @@ ReferenceLookupResult boundResult(ReferenceLookupResult boundRes) {
 
         @Override
         ReferenceLookupResult unboundResult(ReferenceLookupResult boundRes, ReferenceLookupResult unboundRes) {
-            if (boundRes.hasKind(StaticKind.STATIC) &&
+            if (boundRes.isSuccess() && boundRes.sym.isStatic() &&
```
mcimadamore, line 3327: For consistency here I'd prefer to see `boundRes.hasKind(StaticKind.STATIC)` (which is set if the underlying symbol is static). E.g. I think the fix here is really adding the `isSuccess` part, not changing the static check.

mcimadamore: That said - I'm having issue understanding how this has been fixed: if the fix search succeeds and the second search fails to produce a non-static method, then the first search should be preferred. It seems to me that the old code already covers that? Note that `isSuccess` is true whenever the kind != UNDEFINED, so if we check that kind should be STATIC, we already check that is a successful lookup and that the result is static. In other words, I don't see how the new code is different from the old?

mcimadamore (commented):
The debugging logic seems good, but I don't get how the fix is supposed to work, since it seems to replicate the behavior that's already there. I'm probably missing something.

src/jdk.compiler/share/classes/com/sun/tools/javac/comp/Resolve.java
vicente-romero-oracle: I guess that the difference between the new code and the previous one is that in the old code when we check if `boundRes.hasKind(StaticKind.STATIC)` we are checking for a global property of all candidates, meaning that for the previous check to be true, all candidates have to be static. And this is not what the spec is saying, as the spec cares about the staticness, or lack of it, of the most specific method found in each search. Does this make sense?

mcimadamore: Ok - I missed that both symbols are applicable in the first search. So while the most specific symbol is the static one, the StaticKind is set to BOTH, since one is static and the other is not. So, yes, using StaticKind for things like "the first/second search returned a static/non-static method" is incorrect. Your patch fixes that.

I'm dubious as to whether isSuccess is correct though. E.g. isSuccess returns true whenever the StaticKind is != UNDEFINED - but that would also return true in cases where there's an ambiguity in the first search - which I would not describe as a success. Do you agree? If so, can we add a test for that case (e.g. a case where the first search has ambiguous result) and see what happens?

vicente-romero-oracle: I have uploaded a new commit based on your comments and offline discussion, TIA

mcimadamore (approved):
Looks good

mcimadamore (approved):
Looks good

PR #121, opened 2021-01-14
8259641: C2: assert(early->dominates(LCA)) failed: early is high enough

rwestrel:
For the fix of 8258393 that causes this regression, I wrote that the
goal was:

"I propose to be less conservative in anti-dependence analysis for
Phis. For a Phi, when computing the LCA, I think it's sufficient to
only consider region's inputs that we actually reach by following the
memory edges and that's what I propose here."

But that's not what the logic I added does. It checks whether a Phi
input should be taken into account by checking whether it's in the
worklist but the worklist doesn't include mem, the memory state from
which we started.

---------
### Progress

### Issue
 * [JDK-8259641](https://bugs.openjdk.java.net/browse/JDK-8259641): C2: assert(early->dominates(LCA)) failed: early is high enough

### Reviewers

### Download
`$ git fetch https://git.openjdk.java.net/jdk16 pull/121/head:pull/121`
`$ git checkout pull/121`

### DIFF
--- a/test/hotspot/jtreg/compiler/loopopts/TestBrokenAntiDependenceWithPhi.java
+++ b/test/hotspot/jtreg/compiler/loopopts/TestBrokenAntiDependenceWithPhi.java
@@ -1,0 +1,96 @@
+/*
+ * Copyright (c) 2021, Red Hat, Inc. All rights reserved.
+ * DO NOT ALTER OR REMOVE COPYRIGHT NOTICES OR THIS FILE HEADER.
+ *
+ * This code is free software; you can redistribute it and/or modify it
+ * under the terms of the GNU General Public License version 2 only, as
+ * published by the Free Software Foundation.
+ *
+ * This code is distributed in the hope that it will be useful, but WITHOUT
+ * ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or
+ * FITNESS FOR A PARTICULAR PURPOSE.  See the GNU General Public License
+ * version 2 for more details (a copy is included in the LICENSE file that
+ * accompanied this code).
+ *
+ * You should have received a copy of the GNU General Public License version
+ * 2 along with this work; if not, write to the Free Software Foundation,
+ * Inc., 51 Franklin St, Fifth Floor, Boston, MA 02110-1301 USA.
+ *
+ * Please contact Oracle, 500 Oracle Parkway, Redwood Shores, CA 94065 USA
+ * or visit www.oracle.com if you need additional information or have any
+ * questions.
+ */
+
+/**
+ * @test
+ * @bug 8259641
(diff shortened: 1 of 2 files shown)


test/hotspot/jtreg/compiler/loopopts/TestBrokenAntiDependenceWithPhi.java
```diff
@@ -0,0 +1,95 @@
...
+ * questions.
+ */
+
+/**
+ * @test
+ * @bug 8259641
+ * @summary C2: assert(early->dominates(LCA)) failed: early is high enough
+ *
+ * @run main/othervm -Xcomp -XX:CompileOnly=TestBrokenAntiDependenceWithPhi TestBrokenAntiDependenceWithPhi
+ *
```
chhagedorn, line 30: Empty line can be removed

test/hotspot/jtreg/compiler/loopopts/TestBrokenAntiDependenceWithPhi.java
```diff
@@ -0,0 +1,95 @@
...
+ */
+
```
chhagedorn, line 32: You could move the test into the `compiler.loopopts` package.

chhagedorn (approved):
That looks good to me!

TobiHartmann (approved):
Looks good to me.

rwestrel:
Thanks for the reviews, Christian & Tobias.

test/hotspot/jtreg/compiler/loopopts/TestBrokenAntiDependenceWithPhi.java
rwestrel: Which of the lines do you suggest removing? One that starts with a *?

test/hotspot/jtreg/compiler/loopopts/TestBrokenAntiDependenceWithPhi.java
rwestrel: Sure but I usually don't (so they are a several tests without it). Why is it important?

chhagedorn: I've just seen that the majority of the compiler tests are in a package. But I'm not sure if it required. Probably not. So I'm fine with how it is now.

test/hotspot/jtreg/compiler/loopopts/TestBrokenAntiDependenceWithPhi.java
chhagedorn: I've meant the very the last one but it's a minor detail. No need to update it just for that.

PR #122, opened 2021-01-14
8259775: [Vector API] Incorrect code-gen for VectorReinterpret operation

DamonFool:
Hi all,

The code-gen for VectorReinterpret may be wrong on x86.

Let's see the opto-assembly for the reproducer in the JBS, which was actually based on @XiaohongGong 's example in JDK-8259353 and many thanks to her.
```
066     B7: #   out( N1 ) <- in( B6 )  Freq: 0.999994
066     vector_reinterpret_expand XMM0,XMM0     !
066     store_vector [R12 + R11 << 3 + #16] (compressed oop addressing),XMM0
``` 
Please note that the dst and src [1] share the same XMM0 register and movdqu [2] should be generated for this case.
But when dst == src, movdqu actually generates nothing [3], which leads to incorrect result;

For this case, movdqu should not be empty since the upper bits of dst should be zeroed.
The similar error also exists for vmovdqu [4].

I think we should also change movflt [5] to movss but I just can't understand why we have 4-byte vectors.
Isn't the shortest vectors 8-byte on x86?

Thanks.
Best regards,
Jie

[1] https://github.com/openjdk/jdk/blob/master/src/hotspot/cpu/x86/x86.ad#L3354
[2] https://github.com/openjdk/jdk/blob/master/src/hotspot/cpu/x86/x86.ad#L3364
[3] https://github.com/openjdk/jdk/blob/master/src/hotspot/cpu/x86/macroAssembler_x86.cpp#L2490
[4] https://github.com/openjdk/jdk/blob/master/src/hotspot/cpu/x86/macroAssembler_x86.cpp#L2515
[5] https://github.com/openjdk/jdk/blob/master/src/hotspot/cpu/x86/x86.ad#L3379

---------
### Progress

### Issue
 * [JDK-8259775](https://bugs.openjdk.java.net/browse/JDK-8259775): [Vector API] Incorrect code-gen for VectorReinterpret operation

### Reviewers
 * [Rickard Bäckman](https://openjdk.java.net/census#rbackman) (@rickard - **Reviewer**)

### Download
`$ git fetch https://git.openjdk.java.net/jdk16 pull/122/head:pull/122`
`$ git checkout pull/122`


DamonFool:
Hi all,

The reason for the wrong execution is that the upper bits of vector registers fails to be zeroed.
This is because movdqu(XMMRegister dst, XMMRegister src) and vmovdqu(XMMRegister dst, XMMRegister src) were incorrectly optimized when dst == src after JDK-8223347 (Integration of Vector API, Oct 14 20:02:46 2020).
So this seems to be a regression of JDK-8223347.

The 4-byte vectors also be fixed by using movfltz since we are not recommended to use movss directly [1].
And the jtreg test has been added to reproduce this bug on both AVX256 and AVX512 machines.

Could you please review it?

Thanks.
Best regards,
Jie

[1] https://github.com/openjdk/jdk/blob/master/src/hotspot/cpu/x86/macroAssembler_x86.hpp#L1048

rickard (approved):

rickard (approved):

rickard (approved):

rickard (approved):

rickard (approved):

neliasso (approved):
Good to know that the code was introduced in 16 so that no regression is introduced.

Approved.

src/hotspot/cpu/x86/macroAssembler_x86.hpp
```diff
@@ -168,6 +168,9 @@ class MacroAssembler: public Assembler {
   void movflt(XMMRegister dst, AddressLiteral src);
   void movflt(Address dst, XMMRegister src) { movss(dst, src); }
 
+  // Move with zero extension
+  void movfltz(XMMRegister dst, XMMRegister src) { movss(dst, src); }
```
vnkozlov, line 172: Seems `movdbl(XMMRegister dst, XMMRegister src)` has the same issue.

vnkozlov (commented):
Don't forget to request approval for JDK 16 fix integration:
http://openjdk.java.net/jeps/3#Fix-Request-Process

src/hotspot/cpu/x86/macroAssembler_x86.hpp
DamonFool: > Seems `movdbl(XMMRegister dst, XMMRegister src)` has the same issue.

Good catch.
I will try to make a reproducer and fix it in another pr since VectorReinterpret doesn't use it.
Thanks.

DamonFool:
Thanks @rickard , @neliasso and @vnkozlov for your review and comments.

@vnkozlov , jdk16-fix-request has been added in the JBS.
Thanks.

DamonFool:
Will integrate it later since the jdk16-fix-request will be approved after PR is finished.
Thanks.

vnkozlov (approved):
I approved this fix for JDK 16.
I misread your comment and thought you will also fix movdbl() here.
I am fine with fixing it in separate PR.

PR #124, opened 2021-01-15
8259765: ZGC: Handle incorrect processor id reported by the operating system

pliden:
Some environments (e.g. OpenVZ containers) incorrectly report a logical processor id that is higher than the number of processors available. This is problematic, for example, when implementing CPU-local data structures, where the processor id is used to index into an array of length processor_count().

We've received crash reports from Jelastic (a Virtuozzo/OpenVZ user) where they run into this problem. We can workaround the problem in the JVM, until the underlying problem is fixed. Without this workaround ZGC can't be used in this environment.

This is currently a ZGC-specific issue, since ZGC is currently the only part of HotSpot that is using CPU-local data structures, but that could change in the future.

Just to clarify. In a Virtuozzo/OpenZV environment, it seems the underlying problem is not necessarily that sched_getcpu() returns an incorrect processor id, but rather that sysconf(_SC_NPROCESSORS_CONF) returns a too low number. Either way, sched_getcpu() and syconf(_SC_NPROCESSORS_CONF) seems to have different views of the world. This is not an issue in container environments such as Docker.

This patch works around this problem by letting os::processor_id() on Linux detect incorrect processor ids, and convert them to processor id 0. As mentioned in the comment in the code, this is safe, but not optimal for performance if the system actually has more than one processor. There's also a warning printed the first time this happen.

Testing: Manual testing with various fake/incorrect values returned from sched_getcpu().

---------
### Progress

### Issue
 * [JDK-8259765](https://bugs.openjdk.java.net/browse/JDK-8259765): ZGC: Handle incorrect processor id reported by the operating system

### Reviewers
 * [Albert Mingkun Yang](https://openjdk.java.net/census#ayang) (@albertnetymk - Author) ⚠️ Review applies to dbe8bd8910127290c0f0636f5ba6077397aca903
 * [Erik Österlund](https://openjdk.java.net/census#eosterlund) (@fisk - **Reviewer**) ⚠️ Review applies to dbe8bd8910127290c0f0636f5ba6077397aca903

### Download
`$ git fetch https://git.openjdk.java.net/jdk16 pull/124/head:pull/124`
`$ git checkout pull/124`

### DIFF
--- a/src/hotspot/os/linux/os_linux.cpp
+++ b/src/hotspot/os/linux/os_linux.cpp
@@ -4746,16 +4746,16 @@ int os::active_processor_count() {
   return active_cpus;
 }
 
-static volatile int warn_invalid_processor_id = 1;
-
 static bool should_warn_invalid_processor_id() {
   if (os::processor_count() == 1) {
     // Don't warn if we only have one processor
     return false;
   }
 
-  if (Atomic::load(&warn_invalid_processor_id) == 0 ||
-      Atomic::xchg(&warn_invalid_processor_id, 0) == 0) {
+  static volatile int warn_once = 1;
+
+  if (Atomic::load(&warn_once) == 0 ||
+      Atomic::xchg(&warn_once, 0) == 0) {
     // Don't warn more than once
     return false;
   }
@@ -4766,7 +4766,7 @@ static bool should_warn_invalid_processor_id() {
 uint os::processor_id() {
   const int id = Linux::sched_getcpu();
 
-  if (id >= 0 && id < processor_count()) {
+  if (id < processor_count()) {
     return (uint)id;
   }
 
@@ -4781,7 +4781,7 @@ uint os::processor_id() {
     log_warning(os)("Invalid processor id reported by the operating system "
                     "(got processor id %d, valid processor id range is 0-%d)",
                     id, processor_count() - 1);
-    log_warning(os)("Falling back so assuming processor id is 0. "
+    log_warning(os)("Falling back to assuming processor id is 0. "
                     "This could have a negative impact on performance.");
   }
 
--- a/src/hotspot/os/linux/os_linux.cpp
+++ b/src/hotspot/os/linux/os_linux.cpp
@@ -4746,19 +4746,46 @@ int os::active_processor_count() {
   return active_cpus;
 }
 
+static volatile int warn_invalid_processor_id = 1;
+
+static bool should_warn_invalid_processor_id() {
+  if (os::processor_count() == 1) {
+    // Don't warn if we only have one processor
+    return false;
+  }
+
+  if (Atomic::load(&warn_invalid_processor_id) == 0 ||
+      Atomic::xchg(&warn_invalid_processor_id, 0) == 0) {
+    // Don't warn more than once
+    return false;
+  }
+
+  return true;
+}
+
 uint os::processor_id() {
   const int id = Linux::sched_getcpu();
 
-#ifndef PRODUCT
-  if (UseDebuggerErgo1 && id >= _processor_count) {
-    // Some debuggers limit the processor count without limiting
-    // the returned processor ids. Fake the processor id.
-    return 0;
+  if (id >= 0 && id < processor_count()) {
+    return (uint)id;
   }
-#endif
 
-  assert(id >= 0 && id < _processor_count, "Invalid processor id [%d]", id);
-  return (uint)id;
+  // Some environments (e.g. openvz containers and the rr debugger) incorrectly
+  // report a processor id that is higher than the number of processors available.
+  // This is problematic, for example, when implementing CPU-local data structures,
+  // where the processor id is used to index into an array of length processor_count().
+  // If this happens we return 0 here. This is is safe since we always have at least
+  // one processor, but it's not optimal for performance if we're actually executing
+  // in an environment with more than one processor.
+  if (should_warn_invalid_processor_id()) {
+    log_warning(os)("Invalid processor id reported by the operating system "
+                    "(got processor id %d, valid processor id range is 0-%d)",
+                    id, processor_count() - 1);
+    log_warning(os)("Falling back so assuming processor id is 0. "
+                    "This could have a negative impact on performance.");
+  }
+
+  return 0;
 }
 
 void os::set_native_thread_name(const char *name) {


src/hotspot/os/linux/os_linux.cpp
```diff
@@ -4746,19 +4746,46 @@ int os::active_processor_count() {
   return active_cpus;
 }
 
+static volatile int warn_invalid_processor_id = 1;
```
albertnetymk: Maybe moving this var into the function, since it's only used inside it.

albertnetymk (approved):

src/hotspot/os/linux/os_linux.cpp
```diff
@@ -4746,19 +4746,46 @@ int os::active_processor_count() {
...
+  // This is problematic, for example, when implementing CPU-local data structures,
+  // where the processor id is used to index into an array of length processor_count().
+  // If this happens we return 0 here. This is is safe since we always have at least
+  // one processor, but it's not optimal for performance if we're actually executing
+  // in an environment with more than one processor.
+  if (should_warn_invalid_processor_id()) {
+    log_warning(os)("Invalid processor id reported by the operating system "
+                    "(got processor id %d, valid processor id range is 0-%d)",
+                    id, processor_count() - 1);
+    log_warning(os)("Falling back so assuming processor id is 0. "
```
fisk: s/so/to/

src/hotspot/os/linux/os_linux.cpp
```diff
@@ -4746,19 +4746,46 @@ int os::active_processor_count() {
...
+
 uint os::processor_id() {
   const int id = Linux::sched_getcpu();
 
-#ifndef PRODUCT
-  if (UseDebuggerErgo1 && id >= _processor_count) {
-    // Some debuggers limit the processor count without limiting
-    // the returned processor ids. Fake the processor id.
-    return 0;
+  if (id >= 0 && id < processor_count()) {
```
fisk: Do we really need to check if the returned processor ID is negative? That seems a whole new level of environment screwup to me.

fisk (changes requested):

src/hotspot/os/linux/os_linux.cpp
pliden: Doing so will come with the cost of always having to run a pthread_once() in function entry.

src/hotspot/os/linux/os_linux.cpp
pliden: Will fix!

src/hotspot/os/linux/os_linux.cpp
pliden: I'm thinking we should make this safe to call in all cases. God knows what a broken environment might return.

dholmes-ora (commented):
So we have to penalize all correctly functioning users because of one broken environment? Can we not detect this broken environment at startup and inject a workaround then?

Why is this an environment that is important enough that OpenJDK has to make changes to deal with a broken environment?

Cheers,
David

fisk (approved):

fweimer-rh:
What does the affinity mask look like at process startup? It should be possible to look at that and take the maximum CPU ID (plus 1) and `sysconf(_SC_NPROCESSORS_CONF)`. This would be a one-time overhead.

It will not work with container deployments that dynamically alter affinity masks. Are there any?

pliden:
@dholmes-ora 

> So we have to penalize all correctly functioning users because of one broken environment? Can we not detect this broken environment at startup and inject a workaround then?

Not sure what you have in mind here? Having an indirect function call would not result in a lower overhead than the test/branch I've introduced. It's also not necessarily trivial to detect this error at startup, as you would need a reliable way to enumerate all processors (something that seems semi-broken in this environment, which is the root of the problem), bind the current thread to each of them and then check the processor id.

> Why is this an environment that is important enough that OpenJDK has to make changes to deal with a broken environment?

That's of course always judgement call/trade-off. I can't say I have a super good understanding of how common this environment it, but there's at least one "Java cloud provider" that uses this environment.

src/hotspot/os/linux/os_linux.cpp
pliden: After some discussions, we agreed to not to check for negative processor ids.

src/hotspot/os/linux/os_linux.cpp
pliden: I'm was wrong here. Since the initialization if effectively const/constexp, there will not be any "pthread_once" overhead here. Moved the static variable inside the function in the latest commit.

pliden:
It seems there have been e-mails sent that didn't show up here, so I'm answering on GitHub to hopefully re-attach the discussion to this PR.

From the mailing list:
>> Glibc's tst-getcpu.c (which I assume is the test you are referring
>> to?) fails in their environment, so it seems like the affinity mask
>> isn't reliable either.
> 
> What's the nature of the failure?  If it's due to a non-changing
> affinity mask, then using sched_getaffinity data would still be okay.

Glibc's tst-getcpu fails with some version of "getcpu results X should be Y".

There seems to be a disconnect between CPU masks/affinity and what sched_getcpu() returns.

Example (container with 1 CPU):

1. sysconf(_SC_NPROCESSORS_CONF) returns 1
2. sysconf(_SC_NPROCESSORS_ONLN) returns 1
3. sched_getaffinity() returns the mask 00000001
4. sched_setaffinity(00000001) return success, but then sched_getcpu() returns 7(!) Should have returned 0.

Another example (container with 2 CPUs):

1. sysconf(_SC_NPROCESSORS_CONF) returns 2
2. sysconf(_SC_NPROCESSORS_ONLN) returns 2
3. sched_getaffinity() returns the mask 00000011
4. sched_setaffinity(00000001) returns success, but then sched_getcpu() returns 2(!). Should have returned 0.
5. sched_setaffinity(00000010) returns success, but then sched_getcpu() also returns 2(!). Should have returned 1.

It looks like CPUs are virtualized on some level, but not in sched_getcpu(). I'm guessing sched_getcpu() is returning the CPU id of the physical CPU, and not the virtual CPU, or something. So in the last example, maybe both virtual CPUs were scheduled on the same physical CPU.

pliden:
> Does sched_getaffinity actually change the affinity mask?

(assuming you meant sched_setaffinity here...)

You're seem to be right. sched_setaffinity() returns success, but a following call to sched_getaffinity() shows it had no effect.

> I wonder if it just reports a 2**N - 1 unconditionally, with N being the
> number of configured vCPUs for the container.  It probably does that so
> that the population count of the affinity mask matches the vCPU count.
> Likewise for the CPU entries under /sys (currently ignored by glibc
> because of a parser bug) and /proc/stat (the fallback actually used by
> glibc).  There is no virtualization of CPU IDs whatsoever, it looks like
> it's all done to communicate the vCPU count, without taking into account
> how badly this interacts with sched_getcpu.

Yep, that's what it looks like.

pliden:
> So it isn't that sysconf(_SC_NPROCESSORS_CONF) returns a too low number as stated in the PR but rather that after calling sched_setaffinity, sched_getcpu is broken?

It wasn't my intention to claim that sysconf() _is_ the problem here. I just wanted to mention that it _might_ be sysconf() that is the problem. The reason I mentioned that is the because of how Docker behaves. If you give a Docker container 2 CPUs, sysconf() will still return the number of CPUs available on the host system, e.g. 8, and sched_getcpu() will in that case return numbers in the 0-7 range. Of course, this was just an observation, Docker and OpenVZ could do things differently here.

> Either way won't that breakage also potentially affect the NUMA code as well?

We should be good, because libnuma will report that NUMA is not available, so we automatically disable UseNUMA if it's set.

pliden:
@dholmes-ora Do you still have questions or concerns here, or can I go ahead and integrate this?

I've gone through all uses of sysconf(_SC_NPROCESSORS_*) and sched_getaffinity() we have, and they look fine. I've also looked at how the OSContainer stuff behaves in this environment, and it also looks fine. In summary, the only problem I can spot is related to sched_getcpu().

pliden:
Ok, thanks all for reviewing.

/integrate

PR #129, opened 2021-01-22
8260284: C2: assert(_base == Int) failed: Not an Int

chhagedorn:
`PhaseIterGVN::no_dependent_zero_check()` does not handle the case when the divisor `n->in(2)` is top when a Div/Mod node is dying. As a result, we hit the assertion in `is_int()` because the type is top.

The fix is straight forward to handle this case and just return false to avoid any further redundant optimization on a dead node.

Thanks,
Christian

---------
### Progress

### Issue
 * [JDK-8260284](https://bugs.openjdk.java.net/browse/JDK-8260284): C2: assert(_base == Int) failed: Not an Int

### Reviewers

### Download
`$ git fetch https://git.openjdk.java.net/jdk16 pull/129/head:pull/129`
`$ git checkout pull/129`

### DIFF
--- a/src/hotspot/share/opto/phaseX.cpp
+++ b/src/hotspot/share/opto/phaseX.cpp
@@ -1665,12 +1665,20 @@ bool PhaseIterGVN::no_dependent_zero_check(Node* n) const {
     case Op_DivI:
     case Op_ModI: {
       // Type of divisor includes 0?
+      if (n->in(2)->is_top()) {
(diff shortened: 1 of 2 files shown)


TobiHartmann (approved):
Looks good to me.

chhagedorn:
Thank you Tobias for your review!

vnkozlov (approved):
Good.

chhagedorn:
Thanks Vladimir for your review!

PR #130, opened 2021-01-22
8259276: C2: Empty expression stack when reexecuting tableswitch/lookupswitch instructions after deoptimization

iwanowww:
During parsing of `lookupswitch` and `tableswitch` instructions C2 may insert a safepoint. Corresponding JVM state has always re-execute bit set since the interpreter will unconditionally re-execute the instruction after deoptimization is over (see `AbstractInterpreter::bytecode_should_reexecute` for the full list of instructions).  

But `Parse::do_tableswitch()`/`Parse::do_lookupswitch()` attach wrong JVM state: it describes the state **after** the instruction since the first thing they do is they pop the operand from the expression stack. Instead, the JVM state **before** the instruction should be used to respect the re-execution in the interpreter. 

The bug manifests as a stack corruption after deoptimization at a broken safepoint.

The fix is to keep the initial JVM state (before the instruction) intact until all the safepoints at the particular instruction are inserted.   

Testing:
- [ ] hs-tier1 - hs-tier7 (in progress)
- [ ] Kitchensink24h w/ -XX:+DeoptimizeALot (in progress)

---------
### Progress

### Issue
 * [JDK-8259276](https://bugs.openjdk.java.net/browse/JDK-8259276): C2: Empty expression stack when reexecuting tableswitch/lookupswitch instructions after deoptimization

### Reviewers
 * [Dean Long](https://openjdk.java.net/census#dlong) (@dean-long - **Reviewer**)

### Download
`$ git fetch https://git.openjdk.java.net/jdk16 pull/130/head:pull/130`
`$ git checkout pull/130`

### DIFF
--- a/src/hotspot/share/opto/graphKit.cpp
+++ b/src/hotspot/share/opto/graphKit.cpp
@@ -823,8 +823,7 @@ bool GraphKit::dead_locals_are_killed() {
 
 #endif //ASSERT
 
-// Helper function for enforcing certain bytecodes to reexecute if
-// deoptimization happens
+// Helper function for enforcing certain bytecodes to reexecute if deoptimization happens.
 static bool should_reexecute_implied_by_bytecode(JVMState *jvms, bool is_anewarray) {
   ciMethod* cur_method = jvms->method();
   int       cur_bci   = jvms->bci();
@@ -839,8 +838,9 @@ static bool should_reexecute_implied_by_bytecode(JVMState *jvms, bool is_anewarr
     // is limited by dimensions and guarded by flag so in some cases
     // multianewarray() runtime calls will be generated and
     // the bytecode should not be reexecutes (stack will not be reset).
-  } else
+  } else {
     return false;
+  }
 }
 
 // Helper function for adding JVMState and debug information to node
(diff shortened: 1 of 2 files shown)


dean-long (approved):

vnkozlov (approved):
Make sense.

vnkozlov:
Add Fix request for JDK 16 to bug report.

TobiHartmann (approved):
Looks good. Just wondering, although this triggers with stress testing, should we still add a targeted regression test? (Can be done later, since this is for JDK 16).

iwanowww:
Thanks for the reviews, Dean, Vladimir, and Tobias.

> Just wondering, although this triggers with stress testing, should we still add a targeted regression test? (Can be done later, since this is for JDK 16).

The fix introduces an assert which reliably fires during compilation with the problematic test case (Kitchensink). So, I see less value in a dedicated regression test now.

TobiHartmann:
> The fix introduces an assert which reliably fires during compilation with the problematic test case (Kitchensink). So, I see less value in a dedicated regression test now.

Okay, fair enough.

PR #131, opened 2021-01-25
8257215: JFR: Events dropped when streaming over a chunk rotation

egahlin:
Hi, 

Could I have review of an intermittent product issue. It happens about 1 out of 300 the test is run.

By design, chunk files that have been parsed are deleted. Problem is that the filename could be reused, which causes issues. This happens on the server, where the files are read before they are transfered over the network, and on the client when chunks are read from the local repository. 

The logic for creating the chunk file names is now in a separate class (ChunkFilename) so it can be shared by server and client. Another problem was the clock source being used when purging files. It was not always the same as when timestamp was taken from the chunk.

Testing: Ran TestRotate.java 2100 times without failure

Testing: jdk/jdk/jfr

Thanks
Erik

---------
### Progress

### Issue
 * [JDK-8257215](https://bugs.openjdk.java.net/browse/JDK-8257215): JFR: Events dropped when streaming over a chunk rotation

### Reviewers
 * [Markus Grönlund](https://openjdk.java.net/census#mgronlun) (@mgronlun - **Reviewer**)

### Download
`$ git fetch https://git.openjdk.java.net/jdk16 pull/131/head:pull/131`
`$ git checkout pull/131`

### DIFF
--- a/src/jdk.jfr/share/classes/jdk/jfr/internal/PlatformRecording.java
+++ b/src/jdk.jfr/share/classes/jdk/jfr/internal/PlatformRecording.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2016, 2020, Oracle and/or its affiliates. All rights reserved.
+ * Copyright (c) 2016, 2021, Oracle and/or its affiliates. All rights reserved.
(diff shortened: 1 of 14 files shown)


mgronlun (approved):
Looks good.

PR #134, opened 2021-01-26
8260370: C2: LoopLimit node is not eliminated

iwanowww:
As part of JDK-8255000 `C->major_progress` check was replaced with an explicit phase check (`post_loop_opts_phase()`).
But `LoopLimit` was not covered and still relied on `C->major_progress()`. It exposed a problem: sometimes IGVN sets `major_progress` (e.g., see `LoopNode::Ideal()`) and it blocks `LoopLimit` node elimination if the node is processed later (`LoopLimitNode::Ideal()` is misled that loop opts phase is not over).         

The fix migrates `LoopLimitNode::Ideal()` to  `post_loop_opts_phase()`. 

Also, post-loop opts IGVN pass conservatively clears `major_progress()` just in case.  
   
Testing:
- [x] failing test
- [x] hs-tier1 - hs-tier4

---------
### Progress

### Issue
 * [JDK-8260370](https://bugs.openjdk.java.net/browse/JDK-8260370): C2: LoopLimit node is not eliminated

### Reviewers

### Download
`$ git fetch https://git.openjdk.java.net/jdk16 pull/134/head:pull/134`
`$ git checkout pull/134`

### DIFF
--- a/src/hotspot/share/opto/compile.cpp
+++ b/src/hotspot/share/opto/compile.cpp
@@ -1840,6 +1840,8 @@ void Compile::process_for_post_loop_opts_igvn(PhaseIterGVN& igvn) {
 
   C->set_post_loop_opts_phase(); // no more loop opts allowed
 
+  assert(!C->major_progress(), "not cleared");
+
   if (_for_post_loop_igvn.length() > 0) {
     while (_for_post_loop_igvn.length() > 0) {
       Node* n = _for_post_loop_igvn.pop();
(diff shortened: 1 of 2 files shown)


vnkozlov (approved):
Looks good.
Do we have other places where we should replace major_progress()?

iwanowww:
Thanks for the prompt review, Vladimir.

> Do we have other places where we should replace major_progress()?

Yes, there's at least one check left (in `SafePointNode::Identity`). I plan to address it separately.

TobiHartmann (approved):
Looks good to me.

PR #135, opened 2021-01-27
8253353: Crash in C2: guarantee(n != NULL) failed: No Node

vnkozlov:
Nesting loops for this case reaches 400+. But the type of IdeaLoopTree::_nest field is uint8 (the maximum value allowed is 255). The test passed with the following fix proposed by @RealFYang:

```
- uint8_t _nest; // Nesting depth 
+ uint16_t _nest; // Nesting depth
```

I think increase size to `unsigned short` (16 bits) is enough for this value. We will bailout C2 compilation due to big IR graph (MaxNodeLimit = 80000) before we can construct MAX_USHORT (65535)  nested loops. And I added asserts to catch such case if it happens.

New regression test is added.
Passed hs-tier1 - 3 testing.

---------
### Progress

### Issue
 * [JDK-8253353](https://bugs.openjdk.java.net/browse/JDK-8253353): Crash in C2: guarantee(n != NULL) failed: No Node

### Reviewers

### Contributors
 * Fei Yang `<fyang@openjdk.org>`

### Download
`$ git fetch https://git.openjdk.java.net/jdk16 pull/135/head:pull/135`
`$ git checkout pull/135`


iwanowww (approved):
Looks good.

neliasso (approved):
Looks good.

PR #137, opened 2021-01-28
8260608: add a regression test for 8260370

iignatev:
Hi all,

could you please review this small and trivial patch that adds a regression test for [8260370](https://bugs.openjdk.java.net/browse/JDK-8260370) ?

Thanks,
-- Igor

---------
### Progress

### Issue
 * [JDK-8260608](https://bugs.openjdk.java.net/browse/JDK-8260608): add a regression test for 8260370

### Reviewers

### Download
`$ git fetch https://git.openjdk.java.net/jdk16 pull/137/head:pull/137`
`$ git checkout pull/137`

### DIFF
--- a/test/hotspot/jtreg/compiler/loopopts/TestLoopLimitNodeElimination.java
+++ b/test/hotspot/jtreg/compiler/loopopts/TestLoopLimitNodeElimination.java
@@ -1,0 +1,72 @@
+/*
+ * Copyright (c) 2021, Oracle and/or its affiliates. All rights reserved.
+ * DO NOT ALTER OR REMOVE COPYRIGHT NOTICES OR THIS FILE HEADER.
+ *
+ * This code is free software; you can redistribute it and/or modify it
+ * under the terms of the GNU General Public License version 2 only, as
+ * published by the Free Software Foundation.
+ *
+ * This code is distributed in the hope that it will be useful, but WITHOUT
+ * ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or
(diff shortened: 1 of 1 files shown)


test/hotspot/jtreg/compiler/loopopts/TestLoopLimitNodeElimination.java
```diff
@@ -0,0 +1,72 @@
...
+ */
+
+package compiler.loopopts;
+
+public class TestLoopLimitNodeElimination {
+    int a = 400;
+    static int counter = 0;
+    long b[] = new long[a];
+
+    void c(String[] d) {
```
vnkozlov: Please, don't use single latter for constants and method names. It is better to not use them for anything - all names should be meaningful.

Also you need to change RFE to Bug. the process only allows test bugs:
http://openjdk.java.net/jeps/3#rdp-2
`RDP2: P1–P5 doc/test bugs`

iignatev: Vladimir, there is nothing meaningful about these constants, methods, variables — the whole test is randomly generated.

vnkozlov: Got it. But even then, you can at least rename 'c()' to 'test()' and 'a' to 'N'. And add comment that it is randomly generated to avoid complains later.

vnkozlov (approved):
Good.

iignatev:
thanks, Vladimir.

/integrate

PR #139, opened 2021-01-30
8260473: [vector] ZGC: VectorReshape test produces incorrect results with ZGC enabled

casparcwang:
https://bugs.openjdk.java.net/browse/JDK-8260473

Function "PhaseVector::expand_vunbox_node" creates a LoadNode, but forgets to make the LoadNode to pass gc barriers.

Testing: all Vector API related tests have passed.

Original pr: https://github.com/openjdk/jdk/pull/2253

---------
### Progress

### Issue
 * [JDK-8260473](https://bugs.openjdk.java.net/browse/JDK-8260473): [vector] ZGC: VectorReshape test produces incorrect results with ZGC enabled

### Reviewers

### Contributors
 * Stuart Monteith `<smonteith@openjdk.org>`
 * Wang Chao `<casparcwang@tencent.com>`

### Download
`$ git fetch https://git.openjdk.java.net/jdk16 pull/139/head:pull/139`
`$ git checkout pull/139`

### DIFF
--- a/test/hotspot/jtreg/compiler/vectorapi/VectorRebracket128Test.java
+++ b/test/hotspot/jtreg/compiler/vectorapi/VectorRebracket128Test.java
@@ -1,0 +1,162 @@
+/*
+ * Copyright (c) 2021, Oracle and/or its affiliates. All rights reserved.
+ * DO NOT ALTER OR REMOVE COPYRIGHT NOTICES OR THIS FILE HEADER.
+ *
+ * This code is free software; you can redistribute it and/or modify it
+ * under the terms of the GNU General Public License version 2 only, as
+ * published by the Free Software Foundation.
+ *
+ * This code is distributed in the hope that it will be useful, but WITHOUT
+ * ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or
+ * FITNESS FOR A PARTICULAR PURPOSE.  See the GNU General Public License
+ * version 2 for more details (a copy is included in the LICENSE file that
+ * accompanied this code).
+ *
+ * You should have received a copy of the GNU General Public License version
+ * 2 along with this work; if not, write to the Free Software Foundation,
+ * Inc., 51 Franklin St, Fifth Floor, Boston, MA 02110-1301 USA.
+ *
+ * Please contact Oracle, 500 Oracle Parkway, Redwood Shores, CA 94065 USA
+ * or visit www.oracle.com if you need additional information or have any
+ * questions.
+ */
+import jdk.incubator.vector.*;
+import jdk.internal.vm.annotation.ForceInline;
+import org.testng.Assert;
+import org.testng.annotations.Test;
+import org.testng.annotations.DataProvider;
+
+import java.lang.invoke.MethodHandles;
+import java.lang.invoke.VarHandle;
+import java.nio.ByteOrder;
+import java.util.Arrays;
+import java.util.List;
+import java.util.function.IntFunction;
+import java.util.function.IntUnaryOperator;
+import jdk.incubator.vector.VectorShape;
+import jdk.incubator.vector.VectorSpecies;
+import jdk.internal.vm.annotation.ForceInline;
+
+/*
+ * @test
+ * @bug 8260473
+ * @requires vm.gc.Z
+ * @modules jdk.incubator.vector
+ * @modules java.base/jdk.internal.vm.annotation
+ * @run testng/othervm -XX:CompileCommand=compileonly,jdk/incubator/vector/ByteVector.fromByteBuffer
+ *      -XX:-TieredCompilation -XX:CICompilerCount=1 -XX:+UseZGC -Xbatch -Xmx256m VectorRebracket128Test
+ */
+
+@Test
+public class VectorRebracket128Test {
+    static final int INVOC_COUNT = Integer.getInteger("jtreg.compiler.vectorapi.vectorrebracket128test.loop-iterations", 1000);
+    static final int NUM_ITER = 200 * INVOC_COUNT;
+
+    static final VectorSpecies<Integer> ispec128 = IntVector.SPECIES_128;
+    static final VectorSpecies<Float> fspec128 = FloatVector.SPECIES_128;
+    static final VectorSpecies<Long> lspec128 = LongVector.SPECIES_128;
+    static final VectorSpecies<Double> dspec128 = DoubleVector.SPECIES_128;
+    static final VectorSpecies<Byte> bspec128 = ByteVector.SPECIES_128;
+    static final VectorSpecies<Short> sspec128 = ShortVector.SPECIES_128;
+
+    static <T> IntFunction<T> withToString(String s, IntFunction<T> f) {
+        return new IntFunction<T>() {
+            @Override
+            public T apply(int v) {
+                return f.apply(v);
+            }
+
+            @Override
+            public String toString() {
+                return s;
+            }
+        };
+    }
+
+    interface ToByteF {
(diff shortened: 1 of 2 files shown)


test/hotspot/jtreg/compiler/vectorapi/VectorRebracket128Test.java
```diff
@@ -0,0 +1,162 @@
...
+import jdk.internal.vm.annotation.ForceInline;
+
+/*
+ * @test
+ * @bug 8260473
+ * @requires vm.gc.Z
+ * @modules jdk.incubator.vector
+ * @modules java.base/jdk.internal.vm.annotation
+ * @run testng/othervm -XX:CompileCommand=compileonly,jdk/incubator/vector/ByteVector.fromByteBuffer
+ *      -XX:-TieredCompilation -XX:CICompilerCount=1 -XX:+UseZGC -Xbatch -Xmx256m VectorRebracket128Test
```
iwanowww, line 47: Why `-XX:CICompilerCount=1` and `-Xmx256m` are needed?

iwanowww (approved):
Overall, looks good.

test/hotspot/jtreg/compiler/vectorapi/VectorRebracket128Test.java
DamonFool: > Why `-XX:CICompilerCount=1` and `-Xmx256m` are needed?

Thanks @iwanowww for your review.

I discussed the same question with @casparcwang offline.
The reason is:
 - Small heap (-Xmx256m) will help to trigger a gc.
 - compileonly and compilercount=1 will let the VM run slow enough to wait for a gc to be finished.
 
And @casparcwang told me that this bug seems not to be reproduced every time without these JVM args.
Thanks.

neliasso (approved):
Approved. 

Now awaiting release team approval.

test/hotspot/jtreg/compiler/vectorapi/VectorRebracket128Test.java
iwanowww: > compileonly and compilercount=1 will let the VM run slow enough to wait for a gc to be finished.

That's a strange way to provoke the bug. You could just increase the number of iterations instead.

But the right way to fix it is to stress ZGC to continuously run in the background while the test case aggressively unboxes vectors in compiled code. `-Xmx256m` helps with that while `-XX:CICompilerCount=1` is irrelevant.

casparcwang: > > compileonly and compilercount=1 will let the VM run slow enough to wait for a gc to be finished.
> 
> That's a strange way to provoke the bug. You could just increase the number of iterations instead.
> 
> But the right way to fix it is to stress ZGC to continuously run in the background while the test case aggressively unboxes vectors in compiled code. `-Xmx256m` helps with that while `-XX:CICompilerCount=1` is irrelevant.

Yes, it's very weird to provoke the bug like this. If CICompilerCount=1 is removed, the test failed 60% roughly on my machine.
And the iteration has already changed from 100 to 1000,  the run time of the test is nearly 30s on release version of jvm.

If I add the following patch, the test always fails on my machine,

```
diff --git a/test/hotspot/jtreg/compiler/vectorapi/VectorRebracket128Test.java b/test/hotspot/jtreg/compiler/vectorapi/VectorRebracket128Test.java
index 1843ec0..959b29a 100644
--- a/test/hotspot/jtreg/compiler/vectorapi/VectorRebracket128Test.java
+++ b/test/hotspot/jtreg/compiler/vectorapi/VectorRebracket128Test.java
@@ -44,7 +44,7 @@ import jdk.internal.vm.annotation.ForceInline;
  * @modules jdk.incubator.vector
  * @modules java.base/jdk.internal.vm.annotation
  * @run testng/othervm -XX:CompileCommand=compileonly,jdk/incubator/vector/ByteVector.fromByteBuffer
- *      -XX:-TieredCompilation -XX:CICompilerCount=1 -XX:+UseZGC -Xbatch -Xmx256m VectorRebracket128Test
+ *      -XX:-TieredCompilation -XX:+UseZGC -Xmx256m VectorRebracket128Test
  */
 
 @Test
@@ -125,6 +125,14 @@ public class VectorRebracket128Test {
     @ForceInline
     static <E,F>
     void testVectorRebracket(VectorSpecies<E> a, VectorSpecies<F> b, byte[] input, byte[] output) {
+        new Thread(() -> {
+            while (true) {
+                try {
+                    System.gc();
+                    Thread.sleep(100);
+                } catch (Exception e) {}
+            }
+        }).start();
         Vector<E> av = a.fromByteArray(input, 0, ByteOrder.nativeOrder());
         int block;
         assert(input.length == output.length);
```

casparcwang: sorry for the wrong patch above,  the failed reason of the patch above is due to stack creation failure (create 1000 threads). The following is the right stress gc patch.

```
diff --git a/test/hotspot/jtreg/compiler/vectorapi/VectorRebracket128Test.java b/test/hotspot/jtreg/compiler/vectorapi/VectorRebracket128Test.java
index 6b266db..a761ea2 100644
--- a/test/hotspot/jtreg/compiler/vectorapi/VectorRebracket128Test.java
+++ b/test/hotspot/jtreg/compiler/vectorapi/VectorRebracket128Test.java
@@ -44,7 +44,7 @@ import jdk.internal.vm.annotation.ForceInline;
  * @modules jdk.incubator.vector
  * @modules java.base/jdk.internal.vm.annotation
  * @run testng/othervm -XX:CompileCommand=compileonly,jdk/incubator/vector/ByteVector.fromByteBuffer
- *      -XX:-TieredCompilation -XX:CICompilerCount=1 -XX:+UseZGC -Xbatch -Xmx256m VectorRebracket128Test
+ *      -XX:-TieredCompilation -XX:+UseZGC -Xmx256m VectorRebracket128Test
  */
 
 @Test
@@ -59,6 +59,19 @@ public class VectorRebracket128Test {
     static final VectorSpecies<Byte> bspec128 = ByteVector.SPECIES_128;
     static final VectorSpecies<Short> sspec128 = ShortVector.SPECIES_128;
 
+    static {
+        Thread t = new Thread(() -> {
+            while (true) {
+                try {
+                    System.gc();
+                    Thread.sleep(100);
+                } catch (Exception e) {}
+            }
+        });
+        t.setDaemon(true);
+        t.start();
+    }
+
     static <T> IntFunction<T> withToString(String s, IntFunction<T> f) {
         return new IntFunction<T>() {
             @Override
```

iwanowww: Good. Please, file a follow-up RFE to improve the test.

DamonFool: > Good. Please, file a follow-up RFE to improve the test.

OK.
I will help to file a JBS bug once the fix has been merged into the jdk mainline.
It will be only fixed in the jdk17, right?
Thanks.

iwanowww:
> It will be only fixed in the jdk17, right?

Yes, I'm OK with that.

test/hotspot/jtreg/compiler/vectorapi/VectorRebracket128Test.java
DamonFool: > Good. Please, file a follow-up RFE to improve the test.

The RFE has been filed here: https://bugs.openjdk.java.net/browse/JDK-8261152
Thanks.

